PR #13, opened 2024-12-23
actions: add CHANGELOG validation script

sam-berning:
**Issue number:**

Closes #

**Description of changes:**

Adds some rudimentary CHANGELOG validation to the build action that runs on PRs. This should ensure that we don't merge a CHANGELOG that can't be parsed by the release CI.

**Testing done:**

Tested script locally. Against current CHANGELOG:

```bash
$ ./tools/validate-changelog.sh  
CHANGELOG validation PASSED!
```

Against the CHANGELOG with a "v" removed from a version header:

```bash
$ ./tools/validate-changelog.sh
1d0
< 1:# 1.0.2 (2024-12-20)
CHANGELOG validation FAILED! Headers must match the regex '^# v[0-9]\+\.[0-9]\+\.[0-9]\+.'
```

**Terms of contribution:**

### DIFF
--- a/.github/workflows/validate-changelog.yml
+++ b/.github/workflows/validate-changelog.yml
@@ -1,0 +1,13 @@
+name: Validate CHANGELOG
+on:
+  pull_request:
+    branches: [develop]
+    paths:
+      - 'CHANGELOG.md'
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: ./tools/validate-changelog.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -53,4 +53,5 @@ jobs:
       - run: make twoliter check-fmt
       - run: make twoliter check-clippy
       - run: make twoliter check-shell
+      - run: ./tools/validate-changelog.sh
```
ginglis13: this action will ignore changes made solely to markdown files (L13), so if someone were to make a change that only edits the CHANGELOG, this check wouldn't run

sam-berning: oh, good catch, thanks! I'll add a new workflow for validating markdown changes then

cbgbt (approved):

ginglis13 (approved):

rpkelly (approved):

larvacea (approved):

PR #28, opened 2025-01-27
kernels: sort microcode firmware configs

ginglis13:
**Description of changes:**

Sort the list of firmware set in `CONFIG_EXTRA_FIRMWARE` to ensure consistent comparisons between kernel configs when updating a kernel package.

**Testing done:**

Build the kernel-kit from 3f46879 and extract configs before making this change: 

```
tools/collect-kernel-config -o ./kernel_configs_pre_sort
```

And after building with this change

```
tools/collect-kernel-config -o ./kernel_configs_sorted
```

No diff:

```
diff <(grep "CONFIG_EXTRA_FIRMWARE=" ./kernel_configs_pre_sort/config-6.1-x86_64.config) <(grep "CONFIG_EXTRA_FIRMWARE=" ./kernel_configs_sorted/config-6.1-x86_64.config)
```

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/kernel-5.10.spec
+++ b/packages/kernel-5.10/kernel-5.10.spec
@@ -118,7 +118,7 @@ done
 %autopatch -p1
 
 %if "%{_cross_arch}" == "x86_64"
-microcode="$(find %{_cross_libdir}/firmware -type f -path '*/*-ucode/*' -printf '%%P ')"
+microcode="$(find %{_cross_libdir}/firmware -type f -path '*/*-ucode/*' -printf '%%P\n' | sort | tr '\n' ' ')"
 cat <<EOF > ../config-microcode
 CONFIG_EXTRA_FIRMWARE="${microcode}"
 CONFIG_EXTRA_FIRMWARE_DIR="%{_cross_libdir}/firmware"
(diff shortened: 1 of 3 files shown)


packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -118,7 +118,7 @@ done
 %autopatch -p1
 
 %if "%{_cross_arch}" == "x86_64"
-microcode="$(find %{_cross_libdir}/firmware -type f -path '*/*-ucode/*' -printf '%%P ')"
+microcode="$(find %{_cross_libdir}/firmware -type f -path '*/*-ucode/*' -printf '%%P ' | sort)"
```
bcressey: The original `find` command was meant to print all files on the same line, separated by spaces - this won't work for `sort` which operates on lines. So I'd switch to printing files with a newline separator, sort the output, then turn it back into a space-separated string:
```suggestion
microcode="$(find %{_cross_libdir}/firmware -type f -path '*/*-ucode/*' -printf '%%P\n' | sort | tr '\n' ' ')"
```

ginglis13: Yup thanks for the catch 😆, looks like my input was presorted

bcressey (approved):

arnaldo2792 (approved):

PR #29, opened 2025-02-03
Added default config files for xfsprogs

### LINKED ISSUE #4281: XFS Ephemeral Storage fails to mount with: Superblock has unknown incompatible features (0x20) enabled.
**Image I'm using:** `bottlerocket-aws-k8s-1.25-x86_64-v1.26.1-943d9a41`
This is on a `i3en.2xlarge` node with local NVMe flash I'm hoping to use as Ephemeral storage. 
If I set the type to ext4 it works.

**What I expected to happen:** The local flash storage would be formatted and mounted as XFS storage

**What actually happened:** The storage isn't mounted. 

I'm guessing because the xfs tools are newer than the kernel?

Error in dmesg:

```
[  302.626669] XFS (md127): Superblock has unknown incompatible features (0x20) enabled.
[  302.626673] XFS (md127): Filesystem cannot be safely mounted by this kernel.
[  302.626675] XFS (md127): SB validate failed with error -22.
```

**How to reproduce the problem:**

Put this in the config:
```
[settings.bootstrap-commands.k8s-ephemeral-storage]
    commands = [
      ["apiclient", "ephemeral-storage", "init", "-t", "xfs"],
      ["apiclient", "ephemeral-storage" ,"bind", "--dirs", "/var/lib/kubelet"]
    ] 
    essential = true
    mode = "always"
```

Sparksssj:
Closes https://github.com/bottlerocket-os/bottlerocket/issues/4281

**Description of changes:**

Part of https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/295

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/kernel-5.10.spec
+++ b/packages/kernel-5.10/kernel-5.10.spec
@@ -16,6 +16,9 @@ Source100: config-bottlerocket
 Source220: neuron-sysinit.target.drop-in.conf
 Source221: modprobe@neuron.service.drop-in.conf
 
+# XFS configuration
+Source300: mkfs.xfs.conf
+
 # Help out-of-tree module builds run `make prepare` automatically.
 Patch1001: 1001-Makefile-add-prepare-target-for-external-modules.patch
 # Enable INITRAMFS_FORCE config option for our use case.
@@ -52,6 +55,9 @@ Requires: %{name}-devel = %{version}-%{release}
 %if "%{_cross_arch}" == "x86_64"
 Requires: (%{name}-modules-neuron if (%{_cross_os}variant-platform(aws) without %{_cross_os}variant-flavor(nvidia)))
 %endif
+
+# Pull in XFSprogs default config files
+Provides: %{_cross_os}kernel(mkfs-confs) = %{version}-%{release}
 
 # The 5.10 kernel is not FIPS certified.
 Conflicts: %{_cross_os}image-feature(fips)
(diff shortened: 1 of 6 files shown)


packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -53,6 +56,9 @@ Requires: %{name}-devel = %{version}-%{release}
 Requires: (%{name}-modules-neuron if (%{_cross_os}variant-platform(aws) without %{_cross_os}variant-flavor(nvidia)))
 %endif
 
+# Pull in XFSprogs default config files
+Provides: %{_cross_os}kernel(mkfs-confs) = %{version}-%{release}
+
```
bcressey, line 61: You don't need this.
```suggestion
```

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -291,9 +297,13 @@ mkdir -p %{buildroot}%{_cross_unitdir}/modprobe@neuron.service.d
 install -p -m 0644 %{S:221} %{buildroot}%{_cross_unitdir}/modprobe@neuron.service.d/neuron.conf
 %endif
 
+mkdir -p %{buildroot}%{_cross_datadir}/xfs
+install -p -m 0644 %{S:300} %{buildroot}%{_cross_datadir}/xfs/mkfs.xfs.conf
```
bcressey, line 301: Why not just use the corresponding config file that's already in the xfsprogs package?

```suggestion
mkdir -p %{buildroot}%{_cross_datadir}/xfsprogs/mkfs
ln -s lts_5.10.conf default.conf
```

Sparksssj: Nice call, thanks! I didn't realize we can use them in that way.

Sparksssj:
Closed this PR since it is pushed to a wrong branch

PR #43, opened 2025-02-10
Fix 5.15 kernel spec file

### LINKED ISSUE #42: kernel 5.15 not applying patches
https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/40 missed the change to `kernel-5.15.spec` to extract the upstream spec file.

As a result, we end up doing the equivalent of the following during the build:
```
❯ readarray -t patches < <(grep -P "^Patch\d+" missing.spec | sort -n -k1.6 | grep -oP "^Patch\d+: \K.*\.patch$" missing.spec)
grep: missing.spec: No such file or directory
grep: missing.spec: No such file or directory

❯ for patch in ${patches[@]}; do
    patch -p1 <../"$patch"
done

// no patches applied
```

arnaldo2792:
**Issue number:**

Closes #42

**Description of changes:**

In 5b853d34 the 5.15 kernel spec file was not updated to extract the kernel.spec file from the SRPM, preventing the patches from being applied

**Testing done:**

I built the kernel kit without the change and confirmed the list of patches returned by the `grep` command was empty due to the missing `kernel.spec` file. With this change, the list isn't empty and the patches are applied.

For the patch count heuristic, I removed `kernel.spec` from each kernel spec file and confirmed the build fail with the message `Mismatch on patches count!`.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/kernel-5.10.spec
+++ b/packages/kernel-5.10/kernel-5.10.spec
@@ -109,10 +109,18 @@ rpmkeys --checksig %{S:0} --dbpath "${PWD}/rpmdb"
 rm -rf "${PWD}/rpmdb"
 rpm2cpio %{S:0} | cpio -iu linux-%{version}.tar config-%{_cross_arch} "*.patch" kernel.spec
 tar -xof linux-%{version}.tar; rm linux-%{version}.tar
+# Count all the patches extracted from the SRPM
+patches_count=$(find -name "*.patch" | wc -l)
 # Find patch ordering based on the Source0 kernel.spec file from the SRPM.
 # First, find all `PatchNNN` lines. Then, sort by the patch number (-k1.6 in sort sets the 6th char
 # in field 1 of input as the sort parameter). Finally, capture just the patch file name specified.
 readarray -t patches < <(grep -P "^Patch\d+" kernel.spec | sort -n -k1.6 | grep -oP "^Patch\d+: \K.*\.patch$" kernel.spec)
+# Fail the build if there is a mismatch in the number of patches found
+if [[ "${patches_count}" -ne "${#patches[@]}" ]]; then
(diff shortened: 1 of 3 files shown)


ginglis13 (approved):

arnaldo2792:
Forced push adds heuristics to fail the build if there was at least one patch file in the SRPM.

arnaldo2792:
(Force push to remove a commit added by mistake)

rpkelly (approved):

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -109,10 +109,18 @@ rpmkeys --checksig %{S:0} --dbpath "${PWD}/rpmdb"
...
+patches_count=$(find -name "*.patch" | wc -l)
 # Find patch ordering based on the Source0 kernel.spec file from the SRPM.
 # First, find all `PatchNNN` lines. Then, sort by the patch number (-k1.6 in sort sets the 6th char
 # in field 1 of input as the sort parameter). Finally, capture just the patch file name specified.
 readarray -t patches < <(grep -P "^Patch\d+" kernel.spec | sort -n -k1.6 | grep -oP "^Patch\d+: \K.*\.patch$" kernel.spec)
+# Fail the build if there was at least one patch file found but none were found in the kernel.spec
+if [[ "${patches_count}" -gt 0 && "${#patches[@]}" == 0 ]]; then
+  echo "Mismatch on patches count!"
+  exit 1
+fi
```
bcressey: Let's assert more strongly that the number of patches in the array equals the number we extracted.

arnaldo2792:
Forced push includes:
- Fail build if the number of patch files is not the same as the number of patches found.

ginglis13 (approved):

bcressey (approved):

PR #47, opened 2025-02-13
prepare for SDK update

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/244

**Description of changes:**
Update all of the `rpm2cpio | cpio -iu` pipelines to account for the new `rpm2cpio` behavior of prefixing files in the archive with `./`, and let the kernel 6.1 package include a potential `modules.weakdep` file that the new `kmod` creates.

**Testing done:**
Built the kernel kit with and without the newer SDKs.

**Terms of contribution:**

### DIFF
--- a/packages/grub/grub.spec
+++ b/packages/grub/grub.spec
@@ -101,10 +101,10 @@ Summary: Tools for the bootloader with support for Linux and more
 rpmkeys --import %{S:1} --dbpath "${PWD}/rpmdb"
 rpmkeys --checksig %{S:0} --dbpath "${PWD}/rpmdb"
 rm -rf "${PWD}/rpmdb"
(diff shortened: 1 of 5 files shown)


bcressey:
This is admittedly kind of ugly - listing expected files twice, using a glob to match a `modules.weakdep` file that might or might not be there.

I'd originally thought to post this as a draft PR and cut over to the new `rpm2cpio` file patterns as part of an SDK update, and I can still go in that direction. But this works too, and will simplify the SDK update down to just a version bump.

ginglis13 (approved):

arnaldo2792 (approved):

PR #61, opened 2025-03-02
kernel: fix Lustre warnings for GCC 13

bcressey:
**Issue number:**
Related: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/258

**Description of changes:**
Fix two error-level warnings triggered by `-Werror=enum-int-mismatch` with GCC 13.

**Testing done:**
All kernels built successfully.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.15/6001-lustre-fix-Werror-enum-int-mismatch.patch
+++ b/packages/kernel-5.15/6001-lustre-fix-Werror-enum-int-mismatch.patch
@@ -1,0 +1,35 @@
+From 82032af947c7a8b17ec30483a6d4a2d44759561f Mon Sep 17 00:00:00 2001
+From: Ben Cressey <bcressey@amazon.com>
+Date: Tue, 25 Feb 2025 01:25:49 +0000
+Subject: [PATCH] lustre: fix -Werror=enum-int-mismatch
+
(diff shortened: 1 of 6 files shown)


packages/kernel-5.15/6001-lustre-fix-Werror-enum-int-mismatch.patch
```diff
@@ -0,0 +1,35 @@
+From 82032af947c7a8b17ec30483a6d4a2d44759561f Mon Sep 17 00:00:00 2001
```
arnaldo2792, line 1: Is this a patch that should be upstreamed to the Amazon Linux kernel?

arnaldo2792 (approved):

larvacea (approved):
The upstream path as I understand it:
* Amazon Linux
* FSx for Lustre (directly responsible for the patches)
* The Lustre project.

packages/kernel-5.15/6001-lustre-fix-Werror-enum-int-mismatch.patch
bcressey: > Is this a patch that should be upstreamed to the Amazon Linux kernel?

Not sure that they'll care until they move to a newer GCC for kernel builds, but I've mentioned it.

PR #88, opened 2025-04-01
kernels: maintain full bottlerocket kernel configs

dhwaniserai:
**Issue number:**

Closes #

**Description of changes:**

Start maintaining full kernel configs for the kernels in bottlerocket.

**Testing done:**

`tools/collect-microcode-config -o packages/microcode`
`make ARCH=x86_64 && make ARCH=aarch64` pass with the additional config diff check for all configs

**Terms of contribution:**


packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -13,6 +13,8 @@ Source1: gpgkey-99E617FE5DB527C0D8BD5F8E11CF1F95C87F5B1A.asc
 Source2: https://yum.repos.neuron.amazonaws.com/aws-neuronx-dkms-2.19.64.0.noarch.rpm
 Source3: gpgkey-00FA2C1079260870A76D2C285749CAD8646D9185.asc
 Source100: config-bottlerocket
+Source101: config-full-bottlerocket-x86_64
+Source102: config-full-bottlerocket-aarch64
```
bcressey, line 17: For these to be useful, we need to modify the steps below (near `merge_config.sh`) to compare the generated config with this config, and error out if they're not identical.

bcressey (commented):
I'd also really like a script in each kernel's directory that I can use to regenerate the full config file, based on the current source rpm and the `config-bottlerocket` partial config.

Otherwise it will be annoying to make changes like https://github.com/bottlerocket-os/bottlerocket-kernel-kit/commit/40504a02fbab5e537cbae02ee2a4d66015f83571 where only the partial config is changing, not the upstream source rpm.

dhwaniserai:
Rebased kernel_configs to v2.0.1 release and added the merge script and spec file check

packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -146,6 +146,11 @@ scripts/kconfig/merge_config.sh \
 %endif
   %{S:100}
 
+if diff "${KCONFIG_CONFIG}" "%{?_cross_arch:x86_64:%{S:101}:%{S:102}}"; then
+  echo "$diff"
```
ginglis13: the above `diff` command should already print a diff

packages/microcode/config-microcode
```diff
@@ -0,0 +1,2 @@
...1 intel-ucode/06-8e-09 intel-ucode/06-8e-0a intel-ucode/06-8e-0b intel-ucode/06-8e-0c intel-ucode/06-8f-05 intel-ucode/06-8f-06 intel-ucode/06-8f-07 intel-ucode/06-8f-08 intel-ucode/06-96-01 intel-ucode/06-97-02 intel-ucode/06-97-05 intel-ucode/06-9a-03 intel-ucode/06-9a-04 intel-ucode/06-9c-00 intel-ucode/06-9e-09 intel-ucode/06-9e-0a intel-ucode/06-9e-0b intel-ucode/06-9e-0c intel-ucode/06-9e-0d intel-ucode/06-a5-02 intel-ucode/06-a5-03 intel-ucode/06-a5-05 intel-ucode/06-a6-00 intel-ucode/06-a6-01 intel-ucode/06-a7-01 intel-ucode/06-aa-04 intel-ucode/06-b7-01 intel-ucode/06-ba-02 intel-ucode/06-ba-03 intel-ucode/06-ba-08 intel-ucode/06-be-00 intel-ucode/06-bf-02 intel-ucode/06-bf-05 intel-ucode/06-cf-01 intel-ucode/06-cf-02 intel-ucode/0f-00-07 intel-ucode/0f-00-0a intel-ucode/0f-01-02 intel-ucode/0f-02-04 intel-ucode/0f-02-05 intel-ucode/0f-02-06 intel-ucode/0f-02-07 intel-ucode/0f-02-09 intel-ucode/0f-03-02 intel-ucode/0f-03-03 intel-ucode/0f-03-04 intel-ucode/0f-04-01 intel-ucode/0f-04-03 intel-ucode/0f-04-04 intel-ucode/0f-04-07 intel-ucode/0f-04-08 intel-ucode/0f-04-09 intel-ucode/0f-04-0a intel-ucode/0f-06-02 intel-ucode/0f-06-04 intel-ucode/0f-06-05 intel-ucode/0f-06-08 "
```
ginglis13, line 1: could we also add a script for generating this?

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,158 @@
...
+
+check_dependencies() {
+    hash rpm2cpio || usage_error "DEPENDENCY ERROR: Please install rpm2cpio somewhere in your PATH"
+    hash cpio || usage_error "DEPENDENCY ERROR: Please install cpio somewhere in your PATH"
+    hash tar || usage_error "DEPENDENCY ERROR: Please install tar somewhere in your PATH"
+    hash docker || usage_error "DEPENDENCY ERROR: Please install docker somewhere in your PATH"
+}
+
+
+while [[ $# -gt 0 ]]; do
```
ginglis13: let's move arg parsing to the `START MAIN CONTROL FLOW` section

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bash
+
+#
+# Common error handling
+#
+
+exit_trap_cmds=()
+
+on_exit() {
```
ginglis13: This doesn't look like it's used anywhere

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -213,6 +213,11 @@ scripts/kconfig/merge_config.sh \
 %endif
   %{S:100}
 
+if diff "${KCONFIG_CONFIG}" "%{?_cross_arch:x86_64:%{S:101}:%{S:102}}"; then
+  echo "$diff"
```
ginglis13: same as above, shouldn't we be getting this from just performing the diff?

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bash
```
ginglis13, line 1: can you make these scripts executable?

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,158 @@
...
+    if [[ $# -gt 0 ]]; then
+        >&2 echo "Error: $*"
+    fi
+    exit 1
+}
+
+# Function to display usage information
+usage() {
+    cat << EOF
+Usage: $(basename "$0") -v VERSION -r RPM_FILE -s SDK_VERSION
```
ginglis13: The `-v` and `-s` flags are not being parsed or used anywhere

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,158 @@
...
+
+# Verify all required parameters are provided
+if [ -z "$rpm_file" ]; then
+    echo "Error: Missing required parameters"
+    usage
+    exit 1
+fi
+
+# expect $pwd to be packages/kernel-${kver}
+merge_kernel_configs() {
```
ginglis13, line 51: If I interrupt this script, I'm left with a dirty tree. E.g.:

```
On branch develop

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	config-aarch64
	config-full-bottlerocket-aarch64
	config-x86_64
	kernel.spec
	latest-kernel-full-config.sh
	linux-6.1.131/
```

Can we either trap interrupts with a cleanup command, or run this script entirely within the SDK container?

dhwaniserai:
Force push contains changes based on feedback. Added a `collect-microcode-config` script in `tools/`

packages/microcode/config-microcode
dhwaniserai: Added a `collect-microcode-config` script for generating it.

packages/kernel-6.1/latest-kernel-full-config.sh
dhwaniserai: Removed all the dead code in the updated commits

dhwaniserai:
updated commit contains moving the logic for the scripts to a `tmpdir` to avoid any potential conflicts with the working tree

tools/collect-microcode-config
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+output_dir=packages/microcode
```
arnaldo2792: nit: if this is a constant, make it all-caps:

```suggestion
OUTPUT_DIR=packages/microcode
```

packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -146,6 +148,10 @@ scripts/kconfig/merge_config.sh \
 %endif
   %{S:100}
 
+if diff "${KCONFIG_CONFIG}" "%{?_cross_arch:x86_64:%{S:101}:%{S:102}}"; then
+  exit 1
```
arnaldo2792, line 158: Can you `echo` out an error message explaining the failure?

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
```
arnaldo2792, line 1: Somewhere in other PR there was a mention of not using `.sh` at the end of the scripts. Let's stick to that convention.

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+        if [ "${arch}" = "aarch64" ]; then
+            karch="arm64"
+            script_args=("../config-${arch}" "../config-bottlerocket")
+        elif [ "${arch}" = "x86_64" ]; then
+            karch="x86"
+            script_args=("../config-${arch}" "../config-microcode" "../config-bottlerocket")
+        fi
+
+        docker run --rm \
+            -v $br_cfg:/config-bottlerocket \
```
arnaldo2792: There were a couple of violations with `shellcheck`, could you please run it and fix them?

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+    hash docker || usage_error "DEPENDENCY ERROR: Please install docker somewhere in your PATH"
+}
+
+# expect $pwd to be packages/kernel-${kver}
+merge_kernel_configs() {
+    rpm_file=$1
+    kernel_path=$PWD
+
+    version="$(rpm --query --nosignature --queryformat '%{VERSION}' "${rpm_file}")"
+    readarray -t br_patches < <(find "${kernel_path}" -name "*.patch")
```
arnaldo2792, line 56: Do you need to sort this as well? Or is it guaranteed that this `find` will return the patches in the correct order?

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+
+    # Patches from bottlerocket
+    for patch in "${br_patches[@]}"; do
+        patch -p1 <"$patch"
+    done
+
+    popd || bail "Could not move around - 'popd' failed. Lets stop before we break anything further."
+
+    for arch in "x86_64" "aarch64"; do
+
```
arnaldo2792, line 83: nit: extra space not needed

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+        # mount files from the rpm and start the sdk docker container
```
arnaldo2792: this comment should go closer to the `docker` execution.

tools/collect-microcode-config
arnaldo2792: Also, don't forget to run `shellcheck` on this script as well

arnaldo2792 (changes requested):

packages/kernel-5.15/latest-kernel-full-config.sh
dhwaniserai: Yeah the bottlerocket patches are numbered correctly and it is read in the same way in the spec so I wanted to maintain that.

packages/kernel-5.15/latest-kernel-full-config.sh
arnaldo2792: I can't find the PR where it was asked to not use `.sh`. I'll back off from it now, but we should standardize the usage of `.sh` in our scripts because we have some with `.sh` and some without it.

packages/kernel-6.1/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+            -v $microcode_cfg:/config-microcode \
+            -v $al_cfg:/config-${arch} \
+            -v $linux_src:/linux-${version} \
+            -e ARCH=${karch} \
+            -e CROSS_COMPILE=/usr/bin/${arch}-bottlerocket-linux-gnu- \
+            -e KCONFIG_CONFIG=bottlerocket_${arch}_defconfig \
+            -w /linux-${version} \
+            --name ${arch}-config-merger public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.60.0 \
+            ./scripts/kconfig/merge_config.sh \
+            "${script_args[@]}"
```
ginglis13: shellcheck has some complaints about this
```suggestion
        docker run --rm \
            -v "${br_cfg}":/config-bottlerocket \
            -v "${microcode_cfg}":/config-microcode \
            -v "${al_cfg}":"/config-${arch}" \
            -v "${linux_src}":"/linux-${version}" \
            -e ARCH="${karch}" \
            -e CROSS_COMPILE="/usr/bin/${arch}-bottlerocket-linux-gnu-" \
            -e KCONFIG_CONFIG="bottlerocket_${arch}_defconfig" \
            -w "/linux-${version}" \
            --name "${arch}-config-merger" public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.60.0 \
            ./scripts/kconfig/merge_config.sh \
            "${script_args[@]}"
```

tools/collect-microcode-config
dhwaniserai: updated

packages/kernel-5.15/latest-kernel-full-config.sh
dhwaniserai: Made the changes suggested by `shellcheck`

packages/kernel-6.1/latest-kernel-full-config.sh
dhwaniserai: done

packages/kernel-5.15/kernel-5.15.spec
dhwaniserai: Added the config diff check in the `spec`

packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -138,14 +140,17 @@ EOF
...
-KCONFIG_CONFIG="arch/%{_cross_karch}/configs/%{_cross_vendor}_defconfig" \
+export KCONFIG_CONFIG="arch/%{_cross_karch}/configs/%{_cross_vendor}_defconfig"
 scripts/kconfig/merge_config.sh \
   ../config-%{_cross_arch} \
 %if "%{_cross_arch}" == "x86_64"
   ../config-microcode \
 %endif
   %{S:100}
 
+SOURCE_FILE=$([ "%{_cross_arch}" = "x86_64" ] && echo "%{S:101}" || echo "%{S:102}")
```
arnaldo2792:
```suggestion
%if "%{_cross_arch}" == "x86_64"
SOURCE_FILE="%{S:101}"
%else
SOURCE_FILE="%{S:102}"
%endif
```

packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -138,14 +140,17 @@ EOF
...
+diff "${KCONFIG_CONFIG}" "${SOURCE_FILE}" || { echo "error: source and build kernel configurations do not match"; exit 1; }
```
arnaldo2792:
```suggestion
if [[ diff "${KCONFIG_CONFIG}" "${SOURCE_FILE}" ]]; then
   echo "error: source and build kernel configurations do not match"
   exit 1
fi
```

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+            karch="arm64"
+            script_args=("../config-${arch}" "../config-bottlerocket")
+        elif [ "${arch}" = "x86_64" ]; then
+            karch="x86"
+            script_args=("../config-${arch}" "../config-microcode" "../config-bottlerocket")
+        fi
+
+        # mount config files and start the sdk docker container
+        docker run --rm \
+            -v "$br_cfg":/config-bottlerocket \
```
arnaldo2792: nit: just for the sake of consistency, if `shellcheck` doesn't complain about it I'll, use `${}`.

packages/kernel-5.15/latest-kernel-full-config.sh
```diff
@@ -0,0 +1,146 @@
...
+    readarray -t br_patches < <(find "${kernel_path}" -name "*.patch")
+
+    tmpdir=$(mktemp -d)
+    cp "$rpm_file" "$tmpdir/"
+    pushd "$tmpdir" || bail "Could not move around"
+
+    rpm2cpio "${rpm_file}" | cpio -iu {,./}linux-"${version}".tar {,.}config-x86_64 {,.}config-aarch64 {,./}"*.patch" {,./}kernel.spec
+    tar -xof linux-"${version}".tar; rm linux-"${version}".tar
+
+    # Find patch ordering based upstream SRPM and apply in that order
```
arnaldo2792: nit:
```suggestion
    # Find patch ordering based on the upstream SRPM and apply in that order
```

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -205,14 +207,17 @@ EOF
...
+export KCONFIG_CONFIG="arch/%{_cross_karch}/configs/%{_cross_vendor}_defconfig"
 scripts/kconfig/merge_config.sh \
   ../config-%{_cross_arch} \
 %if "%{_cross_arch}" == "x86_64"
   ../config-microcode \
 %endif
   %{S:100}
 
+SOURCE_FILE=$([ "%{_cross_arch}" = "x86_64" ] && echo "%{S:101}" || echo "%{S:102}")
+diff "${KCONFIG_CONFIG}" "${SOURCE_FILE}" || { echo "error: source and build kernel configurations do not match"; exit 1; }
```
arnaldo2792: Same as elsewhere in this PR.

dhwaniserai:
Added changes based on comments ^

packages/kernel-5.15/kernel-5.15.spec
```diff
@@ -138,14 +140,17 @@ EOF
 export ARCH="%{_cross_karch}"
```
ginglis13, line 140: not blocking but might be nice to also diff the generated `../config-microcode` with the maintained `packages/microcode/config-microcode`, although that would likely mean maintaining the microcode config in the kernel- packages instead.

(GH review won't let me select the lines above here to place this comment more specifically)

ginglis13 (commented):
LGTM once @arnaldo2792's feedback is addressed

packages/kernel-5.15/kernel-5.15.spec
dhwaniserai: This case would occur if we want to maintain different `config-microcode` per kernel. Otherwise the overall diff would already print the diff in `config-microcode` if any with the rest of the diffs.

packages/kernel-5.15/kernel-5.15.spec
dhwaniserai: I had made the change in `kernel-6.1` but had missed this one earlier; added it now.

ginglis13 (approved):

tools/collect-microcode-config
```diff
@@ -0,0 +1,57 @@
...
+readonly OUTPUT_DIR
+mkdir -p "${OUTPUT_DIR}"
+
+binaries=()
+cross_libdir="/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib"
+make PACKAGE=microcode ARCH=x86_64
+for microcode_rpm in build/rpms/microcode/*.x86_64.rpm; do
+    while IFS= read -r line; do
+        relative_path="${line#"${cross_libdir}"/firmware/}"
+        binaries+=("$relative_path")
```
arnaldo2792: Did you run `shellcheck` in this script? Didn't it complain about this line? If not, just for consistency, you should use `${}`.

arnaldo2792 (approved):

tools/collect-microcode-config
dhwaniserai: updated

dhwaniserai:
updated syntax from `shellcheck`

arnaldo2792 (approved):

ginglis13 (approved):

PR #93, opened 2025-04-02
kernel: add kernel 6.12

larvacea:
**Description of changes:**

A draft PR to add kernel 6.12 to the kernel kit. If you wish to build the new packages, you must add them to the file `kits/bottlerocket-kernel-kit/Cargo.toml`. If you wish to consume the kernel, the variant Cargo.toml in `bottlerocket` must include the image feature:
```
erofs-root-partition = true
```
This can be added to the file near other image features (such as `xfs-data-partition`).

**Testing done:**

**Terms of contribution:**


packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -85,6 +85,12 @@ config DRM_SIMPLEDRM
+ 	select APERTURE_HELPERS
+ 	select DRM_GEM_SHMEM_HELPER
+ 	select DRM_KMS_HELPER
++	select DRM_DISPLAY_DP_HELPER
++	select DRM_DISPLAY_HELPER
++        select DRM_KMS_HELPER
++	select DRM_SUBALLOC_HELPER
++        select DRM_TTM
++	select DRM_TTM_HELPER
```
bcressey: Why this is a patch as opposed to additional lines in the config file?

Please fix the spacing if the patch is required.

Cargo.toml
```diff
@@ -5,8 +5,10 @@ members = [
     "packages/grub",
     "packages/kernel-5.15",
     "packages/kernel-6.1",
+    "packages/kernel-6.12",
     "packages/kmod-5.15-nvidia",
     "packages/kmod-6.1-nvidia",
+    "packages/kmod-6.12-nvidia",
```
bcressey: Please remove the `kmod-6.12-nvidia` package from this PR and let's handle it in a follow up PR.

It's too much change to review repeatedly if the kernel-devel handling in the 6.12 spec file changes as I expect.

packages/kernel-6.12/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+# Use latest-kernel-srpm-url.sh to get this.
+url = "https://cdn.amazonlinux.com/al2023/blobstore/e48aacb6891d5dc0f536065fa7b52d7b8e7bc0d809fe6f93043b7e8bb0e1fe3b/kernel6.12-6.12.20-23.97.amzn2023.src.rpm"
+sha512 = "18a717b3cb0fa6c19ca6e0567691b3349b92390f8ac6a83555efb7cf026782f20a98292fad10b40f50b2972c39a1d25bddf6c60f49bd98afdc9698e3107498bd"
+force-upstream = true
+
+[[package.metadata.build-package.external-files]]
+# Use latest-neuron-srpm-url.sh to get this.
+url = "https://yum.repos.neuron.amazonaws.com/aws-neuronx-dkms-2.18.12.0.noarch.rpm"
+sha512 = "4ed92e661d0ba368eaf8f60e1a68c202062a26819231fcfd42a5ff05d20ad2f34b82b23359a88e80eea22ee5d0056ad769b6febd5d7e7b161da0e36434ba2579"
+force-upstream = true
```
bcressey: This is outdated, please align it with kernel 6.1.

packages/kernel-6.12/2001-Resolve-static-const-compiler-warnings.patch
```diff
@@ -0,0 +1,151 @@
+From 5d1dae633d16144e6fe21da2e70b179a0c009da3 Mon Sep 17 00:00:00 2001
+From: Martin Harriman <larvacea@mac.com>
+Date: Wed, 5 Mar 2025 20:37:48 +0000
+Subject: [PATCH] Resolve static const compiler warnings
```
bcressey: Why is this needed for 6.12 but not for the other kernels? We use the same GCC and it should be the same Neuron sources.

packages/kernel-6.12/bootconfig-aws.conf
```diff
@@ -0,0 +1,2 @@
+kernel.initcall_blacklist = vmd_drv_init, megasas_init, mpt3sas_init, pqi_init
```
bcressey: This line shouldn't be necessary for either aws or vmware snippets if we're not building the additional metal-only drivers.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,255 @@
+# Because Bottlerocket does not have an initramfs, modules required to mount
+# the root filesystem must be set to y.
+
+# The root filesystem is ext4 or erofs
+CONFIG_EXT4_FS=y
+CONFIG_EROFS_FS=y
+
+# With EROFS, we do not use SQUASHFS
+# CONFIG_SQUASHFS is not set
```
bcressey: Move this down with the other disabled filesystems.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,255 @@
...
+CONFIG_KFENCE=y
+# CONFIG_LEGACY_TIOCSTI is not set
+# CONFIG_LIVEPATCH is not set
+# CONFIG_MAGIC_SYSRQ is not set
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=176
+CONFIG_RANDOM_KMALLOC_CACHES=y
+CONFIG_RANDOMIZE_BASE=y
+CONFIG_RANDOMIZE_KSTACK_OFFSET_DEFAULT=y
+CONFIG_RESET_ATTACK_MITIGATION=y
+CONFIG_X86_USER_SHADOW_STACK=y
```
bcressey: I'd like to pull these out and have them come in a follow-up PR, so it's easier to do bisect builds before and after the changes to check for performance impacts.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,255 @@
...
+
+# Xen blkfront for Xen-based EC2 platforms
+CONFIG_XEN_BLKDEV_FRONTEND=y
+
+# Turn off Xen backend
+# CONFIG_XEN_BACKEND is not set
+
+# Turn off i915
+# CONFIG_DRM_I915 is not set
```
bcressey: Is this still necessary?

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,255 @@
...
+
+# virtio for local testing with QEMU
+CONFIG_VIRTIO=y
+CONFIG_VIRTIO_BLK=y
+CONFIG_VIRTIO_PCI=y
+
+# dm-verity and enabling it on the kernel command line
+CONFIG_BLK_DEV_DM=y
+CONFIG_DAX=y
```
bcressey, line 28: Don't think we need this - `CONFIG_DAX=y` specifically.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,255 @@
...
+CONFIG_DRM_TTM_HELPER=y
+
+# =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+# Disable virtio drivers unused in Bottlerocket
+# CONFIG_DRM_VIRTIO_GPU is not set
+# CONFIG_VIRTIO_DMA_SHARED_BUFFER is not set
+
+# NVDIMM
+# CONFIG_LIBNVDIMM is not set
```
bcressey: Is this needed if `CONFIG_DAX` is not set?

packages/kernel-6.12/latest-kernel-srpm-url.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+cmd='dnf install -q -y --releasever=latest yum-utils && yumdownloader -q --releasever=latest --source --urls kernel6.12'
+docker run --rm amazonlinux:2023 sh -c "${cmd}" \
```
bcressey:
```suggestion
docker run --rm public.ecr.aws/amazonlinux/amazonlinux:2023 sh -c "${cmd}" \
```

packages/kernel-6.12/latest-neuron-srpm-url.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+cmd="
+dnf install -q -y --releasever=latest yum-utils &&
+dnf download -q --repofrompath neuron,https://yum.repos.neuron.amazonaws.com --repo=neuron --urls aws-neuronx-dkms
+"
+docker run --rm amazonlinux:2023 bash -c "${cmd}" \
```
bcressey:
```suggestion
docker run --rm public.ecr.aws/amazonlinux/amazonlinux:2023 bash -c "${cmd}" \
```

packages/microcode/microcode.spec
```diff
@@ -26,11 +26,22 @@ Source3: https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/arc
...
-Requires: %{_cross_os}microcode-intel
+Requires: %{_cross_os}microcode-intel-other
+Requires: %{_cross_os}microcode-intel-ec2
 
 %description
 %{summary}.
 
+%package ec2
+Summary: Microcode for EC2 processors
+URL: https://github.com/bottlerocket-os/bottlerocket/tree/develop/packages/microcode
```
bcressey: This URL is no longer correct and since we're touching the spec we should take the opportunity to fix it.

packages/microcode/microcode.spec
```diff
@@ -26,11 +26,22 @@ Source3: https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/arc
...
+Requires: %{_cross_os}microcode-amd
+Requires: %{_cross_os}microcode-intel-ec2
+
+%description ec2
+%{summary}.
+
+License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-intel-mcu-2018
```
bcressey: This is not the right location for a `License` header`, it should go below `Summary`.

packages/microcode/microcode.spec
```diff
@@ -26,11 +26,22 @@ Source3: https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/arc
...
+Requires: %{_cross_os}microcode-intel-other
+Requires: %{_cross_os}microcode-intel-ec2
 
 %description
 %{summary}.
 
+%package ec2
+Summary: Microcode for EC2 processors
+URL: https://github.com/bottlerocket-os/bottlerocket/tree/develop/packages/microcode
+Requires: %{_cross_os}microcode-amd
```
bcressey: AMD microcode should get the same treatment, while we're making this change. Either:
1. separate ec2 and other subpackages
2. or a single subpackages that provides the ec2 and other capabilities

Then this would become:
```
Requires: %{_cross_os}microcode-amd-ec2
```

And then the dependencies will be more obviously correct to the reader, rather than making them wonder why AMD is different.

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
larvacea: It's a patch because these aren't available as independent configs.

packages/kernel-6.12/2001-Resolve-static-const-compiler-warnings.patch
larvacea: It's probably because I haven't updated the Neuron SRPM. I will fix that and see.

packages/microcode/microcode.spec
larvacea: Let me check the AMD microcode list against the latest AL srpm again. My recollection is that there are no excess AMD microcode files (that is, the non-EC2 list would be empty).

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
larvacea: I will check the status of these Kconfigs, in case they are now menu-visible and can be toggled independently in the new kernel.

packages/kernel-6.12/2001-Resolve-static-const-compiler-warnings.patch
larvacea: The latest srpm still exhibits this behavior:
```
  #11 225.7   149 | const static int root_arch_node_attrs_info_tbl_cnt = sizeof(root_arch_node_attrs_info_tbl) / sizeof(nsysfsmetric_attr_info_t);
  #11 225.7       | ^~~~~
  #11 225.8 /home/builder/rpmbuild/BUILD/bottlerocket-kernel-6.12-6.12.20-build/neuron/v1/neuron_dhal_v1.c:999:1: error: ‘static’ is not at beginning of declaration [-Werror=old-style-declaration]
```
I don't know why gcc would compile the neuron srpm sources differently here than in 6.1, but we do still need to patch the neuron sources here. From the C language specification 6.11 future language directions section:

> The placement of a storage-class specifier other than at the beginning of the declaration specifiers in
a declaration is an obsolescent feature.

I quote this because it was news to me. I am used to the word-salad world where these can come in any order.

packages/microcode/microcode.spec
larvacea: I verified that the "microcode-amd-other" category would be empty. There are five (small) files on the AMD side.

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
larvacea: This is my best guess approximation of the patch we carry in 6.1 to enable the modules we need for GPU drivers in that kernel. I have tried to select the options directly from `config-bottlerocket` without success.

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
```diff
@@ -0,0 +1,29 @@
+From 574ee6cf29164ce587f525434611a76e10e4cc6e Mon Sep 17 00:00:00 2001
+From: Martin Harriman <mharrimn@amazon.com>
+Date: Sun, 30 Mar 2025 22:36:38 +0000
+Subject: [PATCH] Select prerequistes for gpu drivers
```
bcressey: typo, `prerequisites`

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -85,6 +85,12 @@ config DRM_SIMPLEDRM
+ 	select APERTURE_HELPERS
+ 	select DRM_GEM_SHMEM_HELPER
+ 	select DRM_KMS_HELPER
++	select DRM_DISPLAY_DP_HELPER
++	select DRM_DISPLAY_HELPER
++   select DRM_KMS_HELPER
++	select DRM_SUBALLOC_HELPER
++   select DRM_TTM
++	select DRM_TTM_HELPER
```
bcressey: Can we fix the spacing for these or is it somehow correct like this?

packages/kernel-6.12/bootconfig-vmware.conf
```diff
@@ -0,0 +1 @@
+kernel.initcall_blacklist = vmd_drv_init, megasas_init, mpt3sas_init, pqi_init
```
bcressey: This can just be an empty file for VMware.

packages/microcode/microcode.spec
bcressey: I'd still like to see the change I asked for above. Option (2) would avoid an empty package while still letting this change to:
```
Requires: %{_cross_os}microcode-amd-ec2
```

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
bcressey: (note that fixing this should also change the patch file name)

packages/kernel-6.12/2001-Resolve-static-const-compiler-warnings.patch
bcressey: This patch doesn't appear to be applied by the spec file, so it seems like it can be dropped.

bcressey: Let's rename the file so it doesn't have a numbered prefix, something like:
```
neuron-resolve-static-const-compiler-warnings.patch
```

Otherwise it looks like a packaging error where the patch is in the directory but not listed in the spec as `Patch2001`.

packages/kernel-6.12/1006-Select-prerequistes-for-gpu-drivers.patch
larvacea: I can fix the spacing. I can also hate the ASCII tab character. Again. More. As I have for the last 50-ish years.

packages/microcode/microcode.spec
larvacea: Done

larvacea:
Builds fail because of a SHA mismatch with the Neuron srpm. I have updated the draft with the latest Neuron srpm for the 6.12 package.

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1392 @@
...
+Requires: %{_cross_os}microcode-licenses
+%endif
+
+# No bare-metal for this kernel
+Conflicts: %{_cross_os}variant-platform(metal)
+
+# No FIPS submission
+Conflicts: %{_cross_os}image-feature(fips)
+
+
```
bcressey, line 74:
```suggestion
# No squashfs support, rely on erofs for compression
Conflicts: %{_cross_os}image-feature(no-erofs-root-partition)
```

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1392 @@
+%global debug_package %{nil}
```
bcressey, line 1: If you don't disable strip like this, then module signatures end up stripped and modules can't be loaded at runtime.

```suggestion
%global debug_package %{nil}
%global __strip /bin/true
```

packages/microcode/microcode.spec
```diff
@@ -17,27 +17,40 @@ License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-i
...
+License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-intel-mcu-2018
+URL: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/tree/develop/packages/microcode
+Requires: %{_cross_os}microcode-amd-ec2
+Requires: %{_cross_os}microcode-intel-ec2
+
+%description ec2
+%{summary}.
+
+# There is no amd-other subpackage because amd-ec2 includes all AMD microcode.
+%package amd-ec2
```
bcressey:
```suggestion
%package amd
Provides: %{_cross_os}microcode-amd-ec2
Provides: %{_cross_os}microcode-amd-other
```

packages/microcode/microcode.spec
```diff
@@ -17,27 +17,40 @@ License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-i
...
+Requires: %{_cross_os}microcode-amd-ec2
+Requires: %{_cross_os}microcode-intel-other
+Requires: %{_cross_os}microcode-intel-ec2
 
 %description
 %{summary}.
 
-%package amd
-Summary: Microcode for AMD processors
+# Microcode applicable to Intel and AMD processors in EC2
```
bcressey: unnecessary comment if the package summary provides this info:
```suggestion
```

packages/microcode/microcode.spec
```diff
@@ -17,27 +17,40 @@ License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-i
...
+%package ec2
+Summary: Microcode for EC2 AMD and Intel processors
```
bcressey: Rephrase so it doesn't sound like these are "EC2[-specific] AMD processors"
```suggestion
Summary: Microcode for AMD and Intel processors available in EC2
```

packages/microcode/microcode.spec
```diff
@@ -17,27 +17,40 @@ License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-i
...
+URL: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/tree/develop/packages/microcode
+Requires: %{_cross_os}microcode-amd-ec2
+Requires: %{_cross_os}microcode-intel-ec2
+
+%description ec2
+%{summary}.
+
+# There is no amd-other subpackage because amd-ec2 includes all AMD microcode.
+%package amd-ec2
+Summary: Microcode for EC2 AMD processors
```
bcressey:
```suggestion
Summary: Microcode for AMD processors
```

packages/microcode/microcode.spec
```diff
@@ -48,13 +61,23 @@ URL: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git
 %description amd-license
 %{summary}.
 
-%package intel
-Summary: Microcode for Intel processors
+# Intel microcode for Intel processors not present in EC2.
+%package intel-other
+Summary: Microcode for Intel processors (not EC2)
```
bcressey:
```suggestion
Summary: Microcode for Intel processors not available in EC2
```

packages/microcode/microcode.spec
```diff
@@ -48,13 +61,23 @@ URL: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git
 %description amd-license
 %{summary}.
 
-%package intel
-Summary: Microcode for Intel processors
+# Intel microcode for Intel processors not present in EC2.
```
bcressey:
```suggestion
```

packages/microcode/microcode.spec
```diff
@@ -48,13 +61,23 @@ URL: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git
...
 License: LicenseRef-scancode-intel-mcu-2018
 URL: https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files
 Requires: %{_cross_os}microcode-intel-license
 
-%description intel
+%description intel-other
+%{summary}.
+
+%package intel-ec2
+Summary: Microcode for Intel processors on EC2
```
bcressey:
```suggestion
Summary: Microcode for Intel processors available in EC2
```

packages/microcode/microcode.spec
```diff
@@ -116,18 +139,173 @@ install -p -m 0644 intel/intel-ucode/* %{buildroot}%{_cross_libdir}/firmware/int
 
 %files
 
-%files amd
+%files amd-ec2
```
bcressey:
```suggestion
%files amd
```

packages/microcode/microcode.spec
```diff
@@ -17,27 +17,40 @@ License: LicenseRef-scancode-amd-linux-firmware-export AND LicenseRef-scancode-i
...
+
+# There is no amd-other subpackage because amd-ec2 includes all AMD microcode.
+%package amd-ec2
+Summary: Microcode for EC2 AMD processors
 License: LicenseRef-scancode-amd-linux-firmware-export
 URL: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/amd-ucode
 Requires: %{_cross_os}microcode-amd-license
 
-%description amd
+%description amd-ec2
```
bcressey:
```suggestion
%description amd
```

packages/kernel-6.12/config-bottlerocket
larvacea: Leaving this in for the moment, as it changes the build for several dm-verity pieces.

packages/kernel-6.12/config-bottlerocket
larvacea: Apparently it is. Without this, the CONFIG_DRM_I915 driver does build.

arnaldo2792:
> Builds fail because of a SHA mismatch with the Neuron srpm. I have updated the draft with the latest Neuron srpm for the 6.12 package.

Could you please pull this change out of your PR? As this will help with others that are facing the same problem.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,224 @@
+# Because Bottlerocket does not have an initramfs, modules required to mount
+# the root filesystem must be set to y.
+
+# The root filesystem is ext4 or erofs
+CONFIG_EXT4_FS=y
+CONFIG_EROFS_FS=y
+
+
```
bcressey: remove extra newline
```suggestion
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,226 @@
...
+# ARM: EFI_ZBOOT
```
bcressey:
```suggestion
# Compress arm64 kernels
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,226 @@
...
+# CONFIG_SYSFB_SIMPLEFB is not set
+
+# With 6.1 some of the functionalities used by the nvidia driver have moved behind
+# some extra config options CONFIG_DRM_KMS_HELPER and CONFIG_DRM_DISPLAY_HELPER.
+# These config options can not be selected individually, but are selected by certain
+# drivers. Enable the SIMPLEDRM driver, which is a minimal drm driver enabling
+# those helpers for platform provided framebuffers.
+CONFIG_DRM_SIMPLEDRM=y
+CONFIG_DRM_DISPLAY_HELPER=y
+CONFIG_DRM_TTM_HELPER=y
```
bcressey: This can be cleaned up and simplified because of the patch:
```suggestion
# Enable the SIMPLEDRM driver, which is a minimal drm driver that enables the helpers for
# platform provided framebuffers required by the NVIDIA driver.
CONFIG_DRM_SIMPLEDRM=y
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,226 @@
...
+
+# =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+# Disable virtio drivers unused in Bottlerocket
+# CONFIG_DRM_VIRTIO_GPU is not set
+# CONFIG_VIRTIO_DMA_SHARED_BUFFER is not set
+
+# CONFIG_LIBNVDIMM is not set
```
bcressey, line 211: Add a comment explaining why

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1395 @@
...
+
+# Pull in platform-dependent boot config snippets.
+Requires: (%{name}-bootconfig-aws if %{_cross_os}variant-platform(aws))
+Requires: (%{name}-bootconfig-vmware if %{_cross_os}variant-platform(vmware))
+
+# Pull in platform-dependent modules.
+%if "%{_cross_arch}" == "x86_64"
+Requires: (%{name}-modules-neuron if (%{_cross_os}variant-platform(aws) without %{_cross_os}variant-flavor(nvidia)))
+%endif
+
```
bcressey, line 87: add this back since eventually FIPS will be supported:
```suggestion

# Pull in FIPS-related files if needed.
Requires: (%{name}-fips if %{_cross_os}image-feature(fips))

```

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1395 @@
...
+rm -f ../config-* ../*.patch
+
+%if "%{_cross_arch}" == "x86_64"
+cd %{_builddir}
+rpmkeys --import %{S:3} --dbpath "${PWD}/rpmdb"
+rpmkeys --checksig %{S:2} --dbpath "${PWD}/rpmdb"
+rm -rf "${PWD}/rpmdb"
+rpm2cpio %{S:2} | cpio -idmu './usr/src/aws-neuronx-*'
+
```
bcressey:
```suggestion
```

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1395 @@
...
+%{_cross_kmoddir}/modules.dep
+%{_cross_kmoddir}/modules.dep.bin
+%{_cross_kmoddir}/modules.devname
+%{_cross_kmoddir}/modules.order
+%{_cross_kmoddir}/modules.softdep
+%{_cross_kmoddir}/modules.symbols
+%{_cross_kmoddir}/modules.symbols.bin
+%{_cross_kmoddir}/modules.weakdep
+%{_cross_kmoddir}/System.map
+%if "%{_cross_arch}" == "x86_64"
```
bcressey, line 480: restore this newline:
```suggestion

%if "%{_cross_arch}" == "x86_64"
```

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1395 @@
...
+%{_cross_kmoddir}/kernel/arch/x86/crypto/serpent-sse2-x86_64.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/crypto/twofish-avx-x86_64.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/crypto/twofish-x86_64-3way.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/crypto/twofish-x86_64.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/kvm/kvm-amd.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/kvm/kvm-intel.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/kvm/kvm.%{_ko}
+%{_cross_kmoddir}/kernel/arch/x86/platform/intel/iosf_mbi.%{_ko}
+%endif
+
```
bcressey: remove this newline:
```suggestion
```

bcressey (commented):
Some minor notes to clean up the diff between 6.1 and 6.12 specs:
```
git diff --no-index packages/kernel-6.1/kernel-6.1.spec packages/kernel-6.12/kernel-6.12.spec
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -0,0 +1,221 @@
...
+# Disable DAMON subsystem. We currently do not have a good use-case for DAMON.
+# CONFIG_DAMON is not set
+
+# Disable unnecessary framebuffer/drm drivers
+# CONFIG_DRM_BOCHS is not set
+# CONFIG_SYSFB_SIMPLEFB is not set
+
+# Enable the SIMPLEDRM driver, which is a minimal drm driver enabling
+# those helpers for platform provided framebuffers.
+CONFIG_DRM_SIMPLEDRM=m
```
bcressey: This doesn't provide the crucial bit of context from my suggestion, which is that we're doing this for the NVIDIA driver:
```suggestion
# Enable the SIMPLEDRM driver, which is a minimal drm driver that enables the helpers for
# platform provided framebuffers required by the NVIDIA driver.
CONFIG_DRM_SIMPLEDRM=m
```

bcressey (approved):

packages/kernel-6.12/config-bottlerocket
larvacea: Done (but word-wrapped to 80 characters).

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1397 @@
...
+  drop_in="$(printf "%03d\n" "${i}")-${fipsmod}.conf"
+  sed -e "s|__FIPS_MODULE__|${fipsmod}|g" %{S:200} \
+    > %{buildroot}%{_cross_unitdir}/check-fips-modules.service.d/"${drop_in}"
+  (( i+=1 ))
+done
+
+LOWERPATH=$(systemd-escape --path %{_cross_sharedstatedir}/kernel-devel/.overlay/lower)
+mkdir -p %{buildroot}%{_cross_unitdir}/"${LOWERPATH}.mount.d"
+sed -e 's|PREFIX|%{_cross_prefix}|g' %{S:210} \
+  > %{buildroot}%{_cross_unitdir}/"${LOWERPATH}.mount.d"/no-squashfs.conf
```
arnaldo2792, line 389: It took me a bit to piece these together. We should consider moving the kernel-devel mount files from the core kit to the kernel kit. It feels a bit disconnected that we are dropping a config here and expect it to be picked up. If we provide the unit files to configure the kernel sources through this package, it will be less disconnected. Do you have opinions on this @bcressey ?

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1397 @@
...
+
+%description fips
+%{summary}.
+
+%prep
+%if "%{_cross_arch}" == "aarch64"
+%global _cross_kimage vmlinuz.efi
+%endif
+
+%global _ko ko
```
arnaldo2792, line 171: nit: a comment that here that describes why the macro would be nice

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "packages/grub",
     "packages/kernel-5.15",
     "packages/kernel-6.1",
+    "packages/kernel-6.12",
```
arnaldo2792, line 8: I'm posting this message here for the lack of a better place to leave this commit, commits are missing the `signed-off` footer.

packages/kernel-6.12/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+# Use latest-kernel-srpm-url.sh to get this.
+url = "https://cdn.amazonlinux.com/al2023/blobstore/e48aacb6891d5dc0f536065fa7b52d7b8e7bc0d809fe6f93043b7e8bb0e1fe3b/kernel6.12-6.12.20-23.97.amzn2023.src.rpm"
+sha512 = "18a717b3cb0fa6c19ca6e0567691b3349b92390f8ac6a83555efb7cf026782f20a98292fad10b40f50b2972c39a1d25bddf6c60f49bd98afdc9698e3107498bd"
+force-upstream = true
+
+[[package.metadata.build-package.external-files]]
+# Use latest-neuron-srpm-url.sh to get this.
+url = "https://yum.repos.neuron.amazonaws.com/aws-neuronx-dkms-2.20.28.0.noarch.rpm"
+sha512 = "e94ea8244f41fd546fb2aa66a793488069a26724b59a4d05355011fca2643f0874a2971e617563c5c2c402952ed160c3b85548f6013c2c60fbc41c89b39133d8"
+force-upstream = true
```
rpkelly: This should be changed to our mirror.

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -0,0 +1,1397 @@
...
+Version: 6.12.20
+Release: 1%{?dist}
+Summary: The Linux kernel
+License: GPL-2.0 WITH Linux-syscall-note
+URL: https://www.kernel.org/
+# Use latest-kernel-srpm-url.sh to get this.
+Source0: hhttps://cdn.amazonlinux.com/al2023/blobstore/e48aacb6891d5dc0f536065fa7b52d7b8e7bc0d809fe6f93043b7e8bb0e1fe3b/kernel6.12-6.12.20-23.97.amzn2023.src.rpm
+Source1: gpgkey-B21C50FA44A99720EAA72F7FE951904AD832C631.asc
+# Use latest-neuron-srpm-url.sh to get this.
+Source2: https://yum.repos.neuron.amazonaws.com/aws-neuronx-dkms-2.20.28.0.noarch.rpm
```
rpkelly, line 14: See above about using a mirror

larvacea:
Updated Cargo.toml to fetch the Neuron source rpm from the Bottlerocket lookaside cache.

packages/kernel-6.12/neuron-resolve-static-const-compiler-warnings.patch
```diff
@@ -0,0 +1,151 @@
+From 5d1dae633d16144e6fe21da2e70b179a0c009da3 Mon Sep 17 00:00:00 2001
```
ginglis13, line 1: Is there a reason we haven't named this patch prefixed with `1008-` ?

ginglis13 (approved):
Nice!

arnaldo2792 (approved):

packages/kernel-6.12/neuron-resolve-static-const-compiler-warnings.patch
larvacea: Yes: this patch isn't applied when we autopatch the world with the 1000-series patches. It gets applied explicitly as part of unpacking and building the neuron driver.

PR #99, opened 2025-04-15
update development-related packaging for kernel-6.12

bcressey:
**Issue number:**
Related: #74

**Description of changes:**
Remove the various archive subpackages related to kernel development, and replace with a single subpackage that includes the files in way that's easy for downstream packages to depend on.

This prevents the 6.12 kernel from being used to assemble a kmod kit, or to build modules at runtime using a suitable toolchain archive, so conflict with the corresponding Twoliter image feature.

Disable the `kheaders.tar.xz` feature which is not needed for modern tools that rely on BTF. Provide `vmlinux.h` to support ahead-of-time builds for BPF programs targeting a specific kernel kit version.

**Testing done:**
Verified that `/sys/kernel/kheaders.tar.xz` was no longer available at runtime and that the kernel images were around 2 MiB smaller.

Confirmed that `/usr/src/kernels` is still mounted on Bottlerocket hosts, which is currently required by host containers. The directory is writable but empty.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-6.12/config-bottlerocket
+++ b/packages/kernel-6.12/config-bottlerocket
@@ -87,9 +87,6 @@ CONFIG_SECURITY_LANDLOCK=y
 CONFIG_IKCONFIG=y
 CONFIG_IKCONFIG_PROC=y
 
-# kernel headers at /sys/kernel/kheaders.tar.xz
-CONFIG_IKHEADERS=y
-
 # BTF debug info at /sys/kernel/btf/vmlinux
 CONFIG_DEBUG_INFO_BTF=y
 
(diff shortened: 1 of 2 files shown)


rpkelly (approved):

larvacea (approved):

packages/kernel-6.12/kernel-6.12.spec
```diff
@@ -435,20 +408,11 @@ install -p -m 0644 %{S:301} %{buildroot}%{_cross_bootconfigdir}/05-vmware.conf
 %{_cross_includedir}/xen/*
 
 %files devel
-%dir %{_cross_ksrcdir}
+%{_cross_usrsrc}/kernels/%{kmajor}
+%{_cross_ksrcdir}
```
ginglis13, line 412: should this still be `%dir` or do we want to package in everything under `%{_cross_ksrcdir}` for devel?

ginglis13 (approved):

packages/kernel-6.12/kernel-6.12.spec
bcressey: We want to include everything underneath it.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -87,9 +87,6 @@ CONFIG_SECURITY_LANDLOCK=y
 CONFIG_IKCONFIG=y
 CONFIG_IKCONFIG_PROC=y
 
-# kernel headers at /sys/kernel/kheaders.tar.xz
-CONFIG_IKHEADERS=y
```
arnaldo2792, line 91: :partying_face:

arnaldo2792 (approved):

PR #111, opened 2025-04-21
kernel-6.12: set config options for kernel hardening

### LINKED ISSUE #97: Adopt kernel hardening recommendations for 6.12
**What I'd like:**

Selectively add recommendations from kernel-hardening-checker to kernel-6.12. This needs testing both for function (as we have found that some recommended features break bottlerocket) and performance (some will add overhead to kernel calls).

Maintain the status quo.

larvacea:
**Issue number:**
Fixes #97 

**Description of changes:**
Set config options for kernel hardening.

**Testing done:**

Tested in k8s, including very limited benchmarking of kubelet start time.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-6.12/config-bottlerocket
+++ b/packages/kernel-6.12/config-bottlerocket
@@ -48,6 +48,35 @@ CONFIG_FB=y
 CONFIG_FB_EFI=y
 CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 
+# Do not allow writes to a block device if a filesystem is mounted on it.
+# CONFIG_BLK_DEV_WRITE_MOUNTED is not set
+
+# Bottlerocket does not hibernate
+# CONFIG_HIBERNATION is not set
+
+# Do not permit processes to push characters into their controlling tty device
+# CONFIG_LEGACY_TIOCSTI is not set
+
+# Bottlerocket does not livepatch
+# CONFIG_LIVEPATCH is not set
+
+# The "magic sysrq" key is enabled. Set the default operations available
+# through a console device. All operations are always available via
+# /proc/sysrq-trigger, and this default can be overridden.
+# SysRq operations enabled by default:
+# 176 = sync (16) + remount read-only (32) + reboot/poweroff (128)
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=176
+
+# We have a relocatable kernel, so randomize the kernel base address, both
+# physical and virtual.
+CONFIG_RANDOMIZE_BASE=y
+
+# Set the default state of kernel stack address randomization. With
+# randomization on, the kernel stack offset is randomized.
+# The default can be overriden via a kernel command-line option,
+# randomize_kstack_offset=[on/off].
+CONFIG_RANDOMIZE_KSTACK_OFFSET_DEFAULT=y
+
 # Turn off i915
 # CONFIG_DRM_I915 is not set
 


larvacea:
The final list of configurations is:
```
CONFIG_ARCH_MMAP_RND_BITS=33
CONFIG_ARCH_MMAP_RND_COMPAT_BITS=16
# CONFIG_BLK_DEV_WRITE_MOUNTED is not set
# CONFIG_HIBERNATION is not set
# CONFIG_LEGACY_TIOCSTI is not set
CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=176
CONFIG_RANDOMIZE_BASE=y
```
These have been chosen as low-risk, unlikely to cause either a functional or performance problem.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,17 @@ CONFIG_FB=y
 CONFIG_FB_EFI=y
 CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 
+# kernel-hardening-checker recommendations
```
bcressey: Please add more specific comments above each entry, in keeping with existing config changes in this file. Otherwise it sounds like you're blindly applying the recommendations of some other tool.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,17 @@ CONFIG_FB=y
...
+CONFIG_ARCH_MMAP_RND_BITS=33
+CONFIG_ARCH_MMAP_RND_COMPAT_BITS=16
+# CONFIG_BLK_DEV_WRITE_MOUNTED is not set
+# CONFIG_HIBERNATION is not set
```
bcressey, line 55:
```suggestion
# CONFIG_HIBERNATION is not set
# CONFIG_LIVEPATCH is not set
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,17 @@ CONFIG_FB=y
...
+# CONFIG_LEGACY_TIOCSTI is not set
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=176
+CONFIG_RANDOMIZE_BASE=y
+CONFIG_RANDOMIZE_KSTACK_OFFSET_DEFAULT=y
+CONFIG_X86_USER_SHADOW_STACK=y
```
bcressey: As I mentioned offline - this doesn't make sense to enable until userspace binaries are built with CET support.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,17 @@ CONFIG_FB=y
 CONFIG_FB_EFI=y
 CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 
+# kernel-hardening-checker recommendations
+CONFIG_ARCH_MMAP_RND_BITS=33
+CONFIG_ARCH_MMAP_RND_COMPAT_BITS=16
```
bcressey: I'm not inclined to take these two ahead of Amazon Linux; let's work with them so that the config ends up with the best values for each architecture.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,17 @@ CONFIG_FB=y
...
+# CONFIG_BLK_DEV_WRITE_MOUNTED is not set
+# CONFIG_HIBERNATION is not set
+# CONFIG_LEGACY_TIOCSTI is not set
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=176
```
bcressey, line 68: The comment for this one needs to explain what "176" means.

packages/kernel-6.12/config-bottlerocket
larvacea: Please take a look and see if this is the level of detail you wanted for each configuration.

packages/kernel-6.12/config-bottlerocket
larvacea: Done.

packages/kernel-6.12/config-bottlerocket
larvacea: This would allow binaries in a container to take advantage of this CET feature, but certainly we can defer this.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,40 @@ CONFIG_FB=y
 CONFIG_FB_EFI=y
 CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 
+# Config options for kernel hardening.
```
bcressey: This isn't the only set of hardening related options in this config file.
```suggestion
```

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,40 @@ CONFIG_FB=y
...
+# Bottlerocket does not hibernate
+# CONFIG_HIBERNATION is not set
+# Do not permit processes to push characters into
+# their controlling tty device
+# CONFIG_LEGACY_TIOCSTI is not set
+
+# Bottlerocket does not livepatch
+# CONFIG_LIVEPATCH is not set
+
+# The "magic sysrq" key is enabled. Set
```
bcressey: nit: you can use a much longer line length for all of these comments, 80 should be fine.

bcressey:
Please also fix the commit message prefix to `kernel-6.12` so it's clear that it's limited to the new kernel.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -48,6 +48,35 @@ CONFIG_FB=y
...
 CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 
+# Do not allow writes to a block device if a filesystem is mounted on it.
+# CONFIG_BLK_DEV_WRITE_MOUNTED is not set
+
+# Bottlerocket does not hibernate
+# CONFIG_HIBERNATION is not set
+
+# Do not permit processes to push characters into their controlling tty device
+# CONFIG_LEGACY_TIOCSTI is not set
```
bcressey, line 58: Does `apiclient exec` still work with this change?

bcressey (approved):

packages/kernel-6.12/config-bottlerocket
larvacea: Yes, it does.

rpkelly (approved):

PR #114, opened 2025-04-24
build: create latest-kernel-full-config for 6.12

larvacea:
**Description of changes:**

Add a latest-kernel-full-config script, customized for kernel-6.12. This package has a special patch file outside the usual `nnnn-title-of-this-commit.patch series`, and the script must not attempt to apply that patch.

**Testing done:**

Script runs and generates the desired full-config files.

**Terms of contribution:**


bcressey:
Are you planning to enforce the diff check in this PR or as a follow-up?

dhwaniserai:
It needs this diff [check](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/packages/kernel-6.1/kernel-6.1.spec#L223) in the spec. This check also helps in verifying the current generated configs during `make`

larvacea:
Added the current configs to this PR, so we can use the script going forward and make config changes visible in a kernel-update PR.

larvacea:
Found a problem in the script. The script does not generate the correct CONFIG_EXTRA_FIRMWARE configuration correctly. I verified that the package and the resulting AMI has the correct firmware on x86_64, and that the script is doing it wrong.

dhwaniserai:
> Found a problem in the script. The script does not generate the correct CONFIG_EXTRA_FIRMWARE configuration correctly. I verified that the package and the resulting AMI has the correct firmware on x86_64, and that the script is doing it wrong.

The script pulls it over from [here](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/packages/microcode/config-microcode) and using [this](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/tools/collect-microcode-config) helper script.

larvacea:
Fixed:
* Added the file `config-microcode-6-12` to the microcode package (exactly like the file `config-microcode`, but with the shorter list of microcode files shipped in the kernel-6.12 package).
* Used that file to generate the `CONFIG_EXTRA_FIRMWARE` config value on x86_64 (as 6.1 does)
* Added the same "does this match" test in `kernel-6.12.spec` as we have in 6.1.

Test: this builds the current 6.12 package, as expected. Changing the generated config-full-bottlerocket-* files does break the build as it should, so the test is working and the build works only because the configurations match.

dhwaniserai:
Not blocking but can we update this [script](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/tools/collect-microcode-config) or make a new one to help generate config-microcode-6.12 also as we may have to update it at some point?

dhwaniserai (approved):

rpkelly:
> [kernel 6.12: latrst-kernel-full-config script](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/114/commits/0530f47538d3f601b165123f40af7f8fb61496cb)

Nit: Can you please fix the typo in your commit

larvacea:
typo fixed

larvacea:
I have force-pushed the typo fix.

larvacea:
This correctly failed CI because #143 introduced configuration changes. Force-pushed the latest baseline.

dhwaniserai (approved):

arnaldo2792 (approved):

larvacea:
Updated saved configurations to match the latest kernel kit release.

dhwaniserai (approved):
verified that both configs match with latest upstream of 6.12.25

PR #116, opened 2025-04-24
Add Infiniband User MAD

yeazelm:
**Description of changes:**
Some instances need `ib_umad` as a loadable module for full support. This will eventually come in as support from our upstream configuration but this takes the change now while that works through upstream.

**Testing done:**
Built a variant with 6.1 and 6.12 kernels and confirmed the module was loadable:
```
[root@admin]# sudo modprobe ib_umad
[root@admin]# lsmod | grep ib_u
ib_umad                49152  0
ib_core               536576  1 ib_umad
```

**Terms of contribution:**


larvacea (approved):

packages/kernel-6.12/config-bottlerocket
```diff
@@ -216,3 +216,6 @@ CONFIG_DRM_SIMPLEDRM=m
 # CONFIG_SCSI_MPT3SAS is not set
 # CONFIG_SCSI_SMARTPQI is not set
 # CONFIG_SCSI_SAS_ATTRS
+
+# Add ib_umad module
+CONFIG_INFINIBAND_USER_MAD=y
```
bcressey: We don't want this built into the kernel:
```suggestion
# Add ib_umad module
CONFIG_INFINIBAND_USER_MAD=m
```

yeazelm: Whoops, I meant to flip to m before sending this out, I started with y just to confirm it worked. I'll fix it!

packages/kernel-6.1/config-bottlerocket
```diff
@@ -369,3 +369,4 @@ CONFIG_SCSI_SNIC=m
 # CONFIG_DRM_VIRTIO_GPU is not set
 # CONFIG_VIRTIO_DMA_SHARED_BUFFER is not set
 
+CONFIG_INFINIBAND_USER_MAD=y
```
arnaldo2792: You may have missed this? And for consistency, the same comment as in 6.12?

bcressey (approved):

arnaldo2792 (approved):

PR #140, opened 2025-05-09
Cache COPYING file for NVIDIA kmods

arnaldo2792:
**Description of changes:**

Previously, the COPYING file was fetched directly from GitHub as a raw file on each build, leading to occasional failures due to GitHub's rate limiting for unauthenticated requests.

Cache the file between builds to avoid hitting these rate limits during workflow re-executions.

**Testing done:**

:grin:

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -16,6 +16,7 @@ runs:
           build/rpms
           build/state
           target
+          packages/kmod-*-nvidia/COPYING
         key: build-cache
     - run: cargo install cargo-make
       shell: bash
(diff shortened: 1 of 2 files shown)


bcressey:
Why not put `COPYING` in the lookaside cache?

Also FYI, the globs won't match the new r570 kmod packages.

rpkelly:
Also curious why this isn't going into the lookaside cache, but, assuming it's staying in git, don't we need to remove `COPYING` from the `.gitignore`?

arnaldo2792:
Closing in favor of caching in the the lookaside cache.

PR #141, opened 2025-05-12
kmod-6.12-nvidia-r570: install nvoptix.bin from the drivers

### LINKED ISSUE #127: Copy nvoptix.bin from nvidia drivers
**What I'd like:**

Currently, the `nvoptix.bin` file is not being copied from the Nvidia drivers at build time. It is required in order to use the OptiX denoiser. I created a [thread back in October](https://github.com/bottlerocket-os/bottlerocket/discussions/4269) to ask help for mounting this file inside my container after forking the bottlerocket repo but it was apparently not possible with what seemed to be the "legacy stack".

However, it looks like https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/458 fixed it. I'm not 100% sure about it but I tried building the kernel kit before and after this PR got merged with my changed in it and the latter is working as expected:

```
$ find / -name nvoptix.bin
/usr/share/nvidia/nvoptix.bin
```

I created a fork with my changed included [here](https://github.com/emaincourt/bottlerocket-kernel-kit/commit/9c9260534bf29915137e89c8da0f38da528fb310).

I saw that all PRs should address an open issue, so would you be willing to accept this change ?

Thanks

emaincourt:
**Issue number:**

Closes #127

**Description of changes:**

In order to be able to use the Nvidia OptiX denoiser, the `nvoptix.bin` binary should be copied from the drivers at build time so that [`nvidia-container-toolkit` can mount it](https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/internal/discover/graphics.go#L63) inside containers requesting the required graphics capabilities. Help regarding the proper installation location would be highly appreciated as this is my first contribution to Bottlerocket.

**Testing done:**

An AMI with the given changes was built and tested against our EKS clusters for the past 3 weeks. I am definitely willing to run more tests if needed.

**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
+++ b/packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
@@ -278,6 +278,7 @@ install -m 755 nvidia-cuda-mps-control %{buildroot}%{_cross_bindir}
 install -m 755 nvidia-cuda-mps-server %{buildroot}%{_cross_bindir}
 install -m 755 nvidia-persistenced %{buildroot}%{_cross_bindir}
 install -m 4755 nvidia-modprobe %{buildroot}%{_cross_bindir}
+install -m 755 nvoptix.bin %{buildroot}%{_cross_datadir}/nvidia/
 ln -rs %{buildroot}%{_cross_bindir}/nvidia-smi %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/nvidia-smi
 ln -rs %{buildroot}%{_cross_bindir}/nvidia-debugdump %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/nvidia-debugdump
 ln -rs %{buildroot}%{_cross_bindir}/nvidia-cuda-mps-control %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/nvidia-cuda-mps-control
@@ -477,6 +478,7 @@ popd
 %{_cross_libdir}/nvidia/tesla/libnvidia-fbc.so.1
 %{_cross_libdir}/nvidia/tesla/libnvoptix.so.%{tesla_ver}
 %{_cross_libdir}/nvidia/tesla/libnvoptix.so.1
+%{_cross_datadir}/nvidia/nvoptix.bin
 
 # Graphics GLVND libs
 %{_cross_libdir}/nvidia/tesla/libnvidia-glvkspirv.so.%{tesla_ver}


packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -278,6 +278,7 @@ install -m 755 nvidia-cuda-mps-control %{buildroot}%{_cross_bindir}
 install -m 755 nvidia-cuda-mps-server %{buildroot}%{_cross_bindir}
 install -m 755 nvidia-persistenced %{buildroot}%{_cross_bindir}
 install -m 4755 nvidia-modprobe %{buildroot}%{_cross_bindir}
+install -m 755 nvoptix.bin %{buildroot}%{_cross_datadir}/nvidia/
```
arnaldo2792, line 281: For reviewers:

Even though it could be possible [to override] where the NVIDIA Container Runtime/CDI generator looks for this file, the option to override the directories isn't available as a configuration that can be overridden at runtime. The alternatives include `/usr/local/share` or `/etc` so I think `%{_cross_datadir}` should be fine for this file. 

Additionally, moving the file to a different location might break some workloads as the found files will be mounted on the container using the location where they were found in the host. In other words, if we move `nvoptix.bin` to say `/usr/local/share/nvidia` , that's the path that will be used within the container. 

[to override]: https://github.com/NVIDIA/nvidia-container-toolkit/blob/132c9afb6cfa1e6ae0a8f5170566de4b677d2606/internal/runtime/runtime.go#L76

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -477,6 +478,7 @@ popd
 %{_cross_libdir}/nvidia/tesla/libnvidia-fbc.so.1
 %{_cross_libdir}/nvidia/tesla/libnvoptix.so.%{tesla_ver}
 %{_cross_libdir}/nvidia/tesla/libnvoptix.so.1
+%{_cross_datadir}/nvidia/nvoptix.bin
```
arnaldo2792, line 481: Technically not a lib, but part of the `Graphics` mounts:

https://github.com/NVIDIA/nvidia-container-toolkit/blob/adb5e6719de18d0e97bd825693d657d121813166/internal/discover/graphics.go#L52

So its OK to keep it in this section as we originally grouped things depending on the mount group they belonged to.

arnaldo2792 (approved):

rpkelly (approved):

bcressey (approved):

arnaldo2792:
Final test to confirm that the binary is mounted:

```
1388 1374 252:0 /x86_64-bottlerocket-linux-gnu/sys-root/usr/share/nvidia/nvoptix.bin /usr/share/nvidia/nvoptix.bin ro,nosuid,nodev,relatime master:1 - erofs /dev/root ro,seclabel,user_xattr,acl,cache_strategy=readaround
```

Issue #152, opened 2025-05-19
latest-kernel-full-config.sh doesn't work


bcressey:
Not sure what's going on, but the script doesn't work at all for me:
```
❯ ./latest-kernel-full-config.sh -r kernel6.12-6.12.25-32.101.amzn2023.src.rpm
/tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
315426 blocks
/tmp/tmp.acceHDd0S4/linux-6.12.25 /tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
...
patching file usr/Kconfig
/tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
/tmp/tmp.acceHDd0S4/linux-6.12.25 /tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
Using ../config-x86_64 as base
cat: ../config-x86_64: Is a directory
mv: cannot stat 'bottlerocket_x86_64_defconfig': No such file or directory
/tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
/tmp/tmp.acceHDd0S4/linux-6.12.25 /tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
Using ../config-aarch64 as base
cat: ../config-aarch64: Is a directory
mv: cannot stat 'bottlerocket_aarch64_defconfig': No such file or directory
/tmp/tmp.acceHDd0S4 ~/kernelkit/packages/kernel-6.12
~/kernelkit/packages/kernel-6.12
```

The error handling also needs work; the script fails on x86_64 and then proceeds to fail on aarch64 in the same way.

bcressey:
I tried it from the top-level directory and that doesn't work either:
```
❯ ./packages/kernel-6.1/latest-kernel-full-config.sh -r ./packages/kernel-6.1/kernel-6.1.134-152.225.amzn2023.src.rpm
/tmp/tmp.UJqf1Ghhuz ~/kernelkit
rpm2archive: ./packages/kernel-6.1/kernel-6.1.134-152.225.amzn2023.src.rpm: No such file or directory
cpio: premature end of archive
tar: linux-6.1.134.tar.xz: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
rm: cannot remove 'linux-6.1.134.tar.xz': No such file or directory
[...]

PR #157, opened 2025-05-20
Move kernel config script to common location and extract SDK from Twoliter metadata

### LINKED ISSUE #151: avoid embedding SDK versions in helper scripts
The `latest-kernel-full-config.sh` scripts each embed an SDK version, for example [this](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/ab9e58f82a70a913323ce32d2479f503cf6916e7/packages/kernel-6.12/latest-kernel-full-config.sh#L112).

Since the SDK can influence the kernel config, this creates friction when moving between SDK versions, or when testing new SDKs.

As a rule, `Twoliter.toml` is the only file that should have to be touched to update the SDK.

ginglis13:
**Issue number:**

Closes #151, #152

**Description of changes:**

This moves the respective latest-kernel-full-config.sh scripts
from kernel package directories into the common tools/ directory. Logic
is consolidating to include any special casing necessary. The script now
extracts the SDK image from a combination of Twoliter.lock and
Twoliter.override, if an override is present.

Additionally, the script places a dependency on tomlq, a crate for
querying toml files.

Also update READMEs across kernel packages

**Testing done:**

Ran against 3 kernels successfully:

```
./tools/latest-kernel-full-config.sh -r ./packages/kernel-5.15/kernel-5.15.180-123.192.amzn2.src.rpm
```

```
./tools/latest-kernel-full-config.sh -r ./packages/kernel-6.1/kernel-6.1.134-152.225.amzn2023.src.rpm
```

```
./tools/latest-kernel-full-config.sh -r ./packages/kernel-6.12/kernel6.12-6.12.25-32.101.amzn2023.src.rpm
```

Repeated above with an override:

```
# Twoliter.override                                                                                                                                                                                
[bottlerocket.bottlerocket-sdk]
registry = "<redacted>.dkr.ecr.us-west-2.amazonaws.com"
name = "thar-be-a-beta-sdk"
```

**Terms of contribution:**


larvacea (approved):

dhwaniserai (approved):
LGTM

tools/latest-kernel-full-config.sh
```diff
@@ -45,27 +49,61 @@ check_dependencies() {
     hash cpio || usage_error "DEPENDENCY ERROR: Please install cpio somewhere in your PATH"
     hash tar || usage_error "DEPENDENCY ERROR: Please install tar somewhere in your PATH"
     hash docker || usage_error "DEPENDENCY ERROR: Please install docker somewhere in your PATH"
+    hash tq || usage_error "DEPENDENCY ERROR: Please cargo install tomlq somewhere in your PATH"
```
arnaldo2792, line 52: We should update the GH actions to run shell check in our scripts.

arnaldo2792 (approved):

PR #168, opened 2025-05-29
Backport patch to ensure NUL-terminated task->comm buffer

### LINKED ISSUE #167: Kernel panic caused by potential non-terminated string when using cgroupsv1 and perf tracing
**Package I'm using:**

Kernel 6.1

**What I expected to happen:**

The kernel doesn't panic if it encounters a non-terminated string, when cgroupsv1 and perf tracing is enabled.
**What actually happened:**

The kernel panicked due to sporadic non-terminated strings:

```
[25565.764981] Call Trace:
[25565.765206]  <TASK>
[25565.765403]  ? show_trace_log_lvl+0x1c0/0x2d9
[25565.765771]  ? show_trace_log_lvl+0x1c0/0x2d9
[25565.766131]  ? show_trace_log_lvl+0x1c0/0x2d9
[25565.766502]  ? perf_trace_cgroup_migrate.cold+0xc/0x18
[25565.766922]  ? __die_body.cold+0x8/0xd
[25565.767241]  ? die+0x2a/0x50
[25565.767499]  ? do_trap+0xc5/0x110
[25565.767786]  ? fortify_panic+0xf/0x11
[25565.768096]  ? do_error_trap+0x65/0x80
[25565.768413]  ? fortify_panic+0xf/0x11
[25565.768721]  ? exc_invalid_op+0x4c/0x60
[25565.769045]  ? fortify_panic+0xf/0x11
[25565.769354]  ? asm_exc_invalid_op+0x16/0x20
[25565.769826]  ? fortify_panic+0xf/0x11
[25565.770240]  ? fortify_panic+0xf/0x11
[25565.770651]  perf_trace_cgroup_migrate.cold+0xc/0x18
[25565.771151]  cgroup_attach_task+0x1de/0x200
[25565.771596]  __cgroup1_procs_write.isra.0+0xfc/0x150
[25565.772093]  kernfs_fop_write_iter+0x12f/0x1c0
[25565.772552]  vfs_write+0x1f4/0x3b0
[25565.772932]  ksys_write+0x63/0xe0
[25565.773300]  do_syscall_64+0x35/0x80
[25565.773682]  entry_SYSCALL_64_after_hwframe+0x6e/0xd8
```

arnaldo2792:
**Issue number:**

Fixes https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/167

**Description of changes:**

Even though the kernel enforces NUL terminated strings, it seems like sometimes they do appear in the kernel causing it to panic. This has been observed upstream, particularly when reading the `comm` property of a task. There is [a patch] upstream that addresses this problem, but it doesn't apply cleanly, so back port the parts that are of interest to the `strscpy` function.

[a patch]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9022ed0e7e65734d83a0648648589b9fbea8e8c9

**Testing done:**

- I built an AMI with the patch, and confirmed the instance boots. Plus, we have anecdotal data that suggests the panic doesn't occur after this patch is applied.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-6.1/1006-strscpy-write-destination-buffer-only-once.patch
+++ b/packages/kernel-6.1/1006-strscpy-write-destination-buffer-only-once.patch
@@ -1,0 +1,127 @@
+From 9022ed0e7e65734d83a0648648589b9fbea8e8c9 Mon Sep 17 00:00:00 2001
+From: Linus Torvalds <torvalds@linux-foundation.org>
+Date: Sun, 1 Dec 2024 09:23:33 -0800
+Subject: [PATCH] strscpy: write destination buffer only once
+
+The point behind strscpy() was to once and for all avoid all the
+problems with 'strncpy()' and later broken "fixed" versions like
+strlcpy() that just made things worse.
+
+So strscpy not only guarantees NUL-termination (unlike strncpy), it also
+doesn't do unnecessary padding at the destination.  But at the same time
+also avoids byte-at-a-time reads and writes by _allowing_ some extra NUL
+writes - within the size, of course - so that the whole copy can be done
+with word operations.
+
+It is also stable in the face of a mutable source string: it explicitly
+does not read the source buffer multiple times (so an implementation
+using "strnlen()+memcpy()" would be wrong), and does not read the source
+buffer past the size (like the mis-design that is strlcpy does).
+
+Finally, the return value is designed to be simple and unambiguous: if
+the string cannot be copied fully, it returns an actual negative error,
+making error handling clearer and simpler (and the caller already knows
+the size of the buffer).  Otherwise it returns the string length of the
+result.
+
+However, there was one final stability issue that can be important to
+callers: the stability of the destination buffer.
+
+In particular, the same way we shouldn't read the source buffer more
+than once, we should avoid doing multiple writes to the destination
+buffer: first writing a potentially non-terminated string, and then
+terminating it with NUL at the end does not result in a stable result
+buffer.
+
+Yes, it gives the right result in the end, but if the rule for the
+destination buffer was that it is _always_ NUL-terminated even when
+accessed concurrently with updates, the final byte of the buffer needs
+to always _stay_ as a NUL byte.
+
+[ Note that "final byte is NUL" here is literally about the final byte
+  in the destination array, not the terminating NUL at the end of the
+  string itself. There is no attempt to try to make concurrent reads and
+  writes give any kind of consistent string length or contents, but we
+  do want to guarantee that there is always at least that final
+  terminating NUL character at the end of the destination array if it
+  existed before ]
+
+This is relevant in the kernel for the tsk->comm[] array, for example.
+Even without locking (for either readers or writers), we want to know
+that while the buffer contents may be garbled, it is always a valid C
+string and always has a NUL character at 'comm[TASK_COMM_LEN-1]' (and
+never has any "out of thin air" data).
+
+So avoid any "copy possibly non-terminated string, and terminate later"
+behavior, and write the destination buffer only once.
+
+Signed-off-by: Linus Torvalds <torvalds@linux-foundation.org>
+[vyaghras:
+  - backport to 6.1
+  - apply changes in original patch to strscpy, since sized_strscpy
+    doesn't exist in 6.1
+]
+Signed-off-by: Shikha Vyaghra <vyaghras@amazon.com>
+Signed-off-by: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+---
+ lib/string.c | 24 ++++++++++++++++++------
+ 1 file changed, 18 insertions(+), 6 deletions(-)
+
+diff --git a/lib/string.c b/lib/string.c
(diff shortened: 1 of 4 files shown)


larvacea (approved):

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -46,6 +46,8 @@ Patch1004: 1004-af_unix-increase-default-max_dgram_qlen-to-512.patch
 # Drop AL revert of upstream patch to minimize delta. The necessary dependency
 # options for nvidia are instead included through DRM_SIMPLE
 Patch1005: 1005-Revert-Revert-drm-fb_helper-improve-CONFIG_FB-depend.patch
+# Prevent copying possible non-terminated strings
```
bcressey: While the patch has wider ramifications because of the many uses of `strscpy`, the specific motivation is to ensure NUL-termination for `task->comm`:
```suggestion
# Backport patch to ensure NUL-terminated task->comm buffer
```

bcressey (approved):
This should be submitted upstream, let's figure out how to do that.

arnaldo2792:
Forced push includes:
- Address feedback in 6.1 spec file
- Backport the original commit for the 6.12 kernel

packages/kernel-6.12/1007-strscpy-write-destination-buffer-only-once.patch
```diff
@@ -0,0 +1,119 @@
...
+This is relevant in the kernel for the tsk->comm[] array, for example.
+Even without locking (for either readers or writers), we want to know
+that while the buffer contents may be garbled, it is always a valid C
+string and always has a NUL character at 'comm[TASK_COMM_LEN-1]' (and
+never has any "out of thin air" data).
+
+So avoid any "copy possibly non-terminated string, and terminate later"
+behavior, and write the destination buffer only once.
+
+Signed-off-by: Linus Torvalds <torvalds@linux-foundation.org>
```
bcressey, line 60: The convention for this kind of backport appears to be:
```
commit <xyz> upstream.

<original commit message>

<original sign-off chain>
<your sign-off>
```

See [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.12.31&id=ecb9d3123bef7307a5ad16d05ce65fe9d9cf8afb) for an example.

So the patch would look like:
```
commit 9022ed0e7e65734d83a0648648589b9fbea8e8c9 upstream.

The point behind strscpy() was to once and for all avoid all the
problems with 'strncpy()' and later broken "fixed" versions like
strlcpy() that just made things worse.
...

So avoid any "copy possibly non-terminated string, and terminate later"
behavior, and write the destination buffer only once.

Signed-off-by: Linus Torvalds <torvalds@linux-foundation.org>
Signed-off-by: Arnaldo Garcia Rincon <agarrcia@amazon.com>
```

For the 6.1 patch, you could do the same thing but also add Shikha's sign-off.

[This commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.12.31&id=3846e2bea565ee1c5195dcc625fda9868fb0e3b3) gives an example of how a `[ note describing backport changes.]` can be formatted. If present (as it would be for the 6.1 patch) it should come before the new sign-offs.

packages/kernel-6.12/1007-strscpy-write-destination-buffer-only-once.patch
```diff
@@ -0,0 +1,119 @@
+From b7d33b4621b8e722f69df30443421712c706bf6c Mon Sep 17 00:00:00 2001
```
bcressey:
```suggestion
From 9022ed0e7e65734d83a0648648589b9fbea8e8c9 Mon Sep 17 00:00:00 2001
```

sky1122 (approved):

arnaldo2792:
Forced push addresses:
- Commit ID matches the upstream commit
- The patch is formatted as suggested

arnaldo2792:
(Forced push fixes the patch format for the 6.1 kernel per suggestions)

sky1122 (approved):

packages/kernel-6.1/1006-strscpy-write-destination-buffer-only-once.patch
```diff
@@ -0,0 +1,128 @@
...
+never has any "out of thin air" data).
+
+So avoid any "copy possibly non-terminated string, and terminate later"
+behavior, and write the destination buffer only once.
+
+Signed-off-by: Linus Torvalds <torvalds@linux-foundation.org>
+[vyaghras:
+  - backport to 6.1
+  - apply changes in original patch to strscpy, since sized_strscpy
+    doesn't exist in 6.1]
```
bcressey, line 64: Per the example commit I linked to, and my earlier offline suggestion, it might be more conventional to write this note like this:
```
[vyaghras: adjust context since e6584c3964f2f ("string: Allow 2-argument
strscpy()") is not present]
```

Issue #173, opened 2025-05-30
tools/latest-kernel-full-config.sh does not work for any UID other than 1000


ginglis13:
**Platform I'm building on:**

x86_64, fedora instance

**What I expected to happen:**

When running the script with an effective UID other than `1000`, I can generate Bottlerocket kernel configs

**What actually happened:**

I have a new user with UID 1002. I see permission denied errors when running the script:

```
$ id
uid=1002(ginglis) gid=1002(ginglis) groups=1002(ginglis),4(adm),10(wheel),190(systemd-journal),993(docker) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
$ ./tools/latest-kernel-full-config.sh -r ./packages/kernel-5.15/kernel-5.15.182-123.190.amzn2.src.rpm
...
/tmp/tmp.oJCtvtUFQK ~/bottlerocket-kernel-kit/packages/kernel-5.15 ~/bottlerocket-kernel-kit/packages ~/bottlerocket-kernel-kit
/tmp/tmp.oJCtvtUFQK/linux-5.15.182 /tmp/tmp.oJCtvtUFQK ~/bottlerocket-kernel-kit/packages/kernel-5.15 ~/bottlerocket-kernel-kit/packages ~/bottlerocket-kernel-kit
mktemp: failed to create file via template './.tmp.config.XXXXXXXXXX': Permission denied
```

**How to reproduce the problem:**

Create a new user other than the default user. Run `./tools/latest-kernel-full-config.sh` against some SRPM, expect above failure.

PR #182, opened 2025-06-05
tools: run bulk of full-config in container

### LINKED ISSUE #174: Modify tools/latest-kernel-full-config.sh to run in the bottlerocket-SDK container
**What I'd like:**

Keep the status quo

larvacea:
Move all of the work to patch, merge, and generate configurations into an sdk container. The previous script only ran the merge-config.sh script in the container.

**Issue number:**

Closes #174

**Description of changes:**

Move the work to extract, patch, and merge kernel configuations into a new inner-full-config script, and execute that script in an SDK container. As seen in issue #173, running this logic on a build machine can break for many reasons. Running inside the SDK world gives the script a stable environment.

**Testing done:**

Generated configurations for all three kernels. Verified that the resulting files were in fact rebuilt, but were identical to the files checked in to GitHub.

Both scripts are shellcheck clean.

**Terms of contribution:**

### DIFF
--- a/tools/latest-kernel-full-config.sh
+++ b/tools/latest-kernel-full-config.sh
@@ -76,89 +76,18 @@ merge_kernel_configs() {
     sdk_image=$2
     kernel_path=$PWD
 
-    version="$(rpm --query --nosignature --queryformat '%{VERSION}' "${rpm_file}")"
-    majorminor=${version%.*} # Trim after last '.', e.g. 6.1.132 -> 6.1
-    if [ "${majorminor}" == "6.12" ]; then
-        # kernel 6.12 has a patch file that is not applied to the kernel sources, so
-        # only pick up the 1000-series kernel patches for our purposes here.
-        readarray -t br_patches < <(find "${kernel_path}" -maxdepth 1 -name "10*.patch")
-        spec_file="kernel6.12.spec"
-        microcode_file="config-microcode-6-12"
-    else
-        readarray -t br_patches < <(find "${kernel_path}" -maxdepth 1 -name "*.patch")
-        spec_file="kernel.spec"
-        microcode_file="config-microcode"
-    fi
 
     tmpdir=$(mktemp -d)
-    cp "${rpm_file}" "${tmpdir}/"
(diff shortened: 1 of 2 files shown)


tools/latest-kernel-full-config.sh
```diff
@@ -76,89 +76,18 @@ merge_kernel_configs() {
...
-
-    popd || bail "Could not move around - 'popd' failed. Lets stop before we break anything further."
-
+    cp "${rpm_file}" "${tmpdir}/kernel-source.rpm"
+
+    docker run --rm \
+        -v "${kernel_path}/":/kernel-package \
+        -v "${tmpdir}":/work \
+        -v "${kernel_path}/../../":/package \
+        -u "$(id -u)":1000 \
```
dhwaniserai, line 87: Can we remove the use of `uid` here?

tools/latest-kernel-full-config.sh
```diff
@@ -76,89 +76,18 @@ merge_kernel_configs() {
...
-            ./scripts/kconfig/merge_config.sh \
-            "${script_args[@]}"
-
-        mv -f "bottlerocket_${arch}_defconfig" "${kernel_path}/config-full-bottlerocket-${arch}"
-        popd || bail "Could not move around - 'popd' failed. Lets stop before we break anything further."
-    done
-
-    popd || bail "Could not move around - 'popd' failed. Lets stop before we break anything further."
-
+    cp "${rpm_file}" "${tmpdir}/kernel-source.rpm"
```
arnaldo2792, line 81: dumb question, why don't we make this a target of the Makefile, and then we can keep `latest-kernel-full-config.sh` as the name of the script?. The parameters to the Makefile target will be the same as with the script.

tools/latest-kernel-full-config.sh
arnaldo2792: We should mimic what happens here:

https://github.com/bottlerocket-os/twoliter/blob/d00e650346a50da9fb25a521fc56edb1cd8562e7/twoliter/embedded/docker-go#L87

sky1122:
close this since we already have follow up PRs

PR #204, opened 2025-07-08
Add IMEX for 6.12 NVIDIA R570 driver

yeazelm:
**Description of changes:**
This adds IMEX as a subpackage that provides support for GB200 devices which require the IMEX process to coordinate efficient communication between devices in ComputeDomains. The DRA device plugin will use these binaries when creating channels but GB200 hardware is required. This DRA DaemonSet provides [its own configuration file](https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/4f10463592f4c877b0c252966d1b0ee14f3774a1/templates/compute-domain-daemon-config.tmpl.cfg) so only the binaries are provided. We do not provide a systemd service file either because the intention is to let the DRA driver manage this.

**Testing done:**
Built an `aws-k8s-1.33-nvidia` with this kit and confirmed the binaries run. 
```
bash-5.1# nvidia-imex --version
IMEX version is: 570.148.08
bash-5.1# nvidia-imex-ctl
Unable to open configuration file at /etc/nvidia-imex/config.cfg
Usage: nvidia-imex-ctl [-i <IP> <PORT>|-u <unix domain socket path>|-c <nvidia-imex config.cfg path>] <-n|-N|-q|-s|-d|-m> [-t <timeout>]
Where:
       -i <IP> <PORT>: Connect to the IMEX daemon's control service at the specified ip address and port.
       -u <unix domain socket path>: Connect to the IMEX daemon's control service at the specified unix domain socket path.
       -c <nvidia-imex config path>: Path to nvidia-imex configuration file.  Default value is /etc/nvidia-imex/config.cfg
NOTE:  If none of -i, -u, -c are provided, -c is assumed with default value.
       -n: Continuously monitor the entire IMEX domain status.  Requires config file (default or via -c)
       -N: Gets the full status of the entire IMEX domain.  Requires config file (default or via -c)
       -q: Query the status of the IMEX daemon once and return.
       -s: Subscribe to the status of the IMEX daemon.  Can be used with -t to terminate after a certain amount of time.
       -d: Dump performance metrics of the IMEX daemon once and return.
       -m: Subscribe to performance metrics of the IMEX daemon.  Can be used with -t to terminate after a certain amount of time.
       -t: Timeout (in milliseconds) for the -s command to terminate.  Unspecified means it will run until IMEX terminates.
       -r: Terminate status subscription when IMEX daemon returns READY status.
       -a: Only used with -q and -c.  Query status of all nodes configured, based on the config file in -c (or default.
       -j: Output of the -n/-N command will be in JSON, rather than formatted.
NOTE: only one of -i, -u, or -c can be specified.  If none are specified, -c is assumed with the default nvidia-imex config file location.
      also, only one command can be executed.
```

**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.12-nvidia-r570/Cargo.toml
+++ b/packages/kmod-6.12-nvidia-r570/Cargo.toml
@@ -37,6 +37,16 @@ sha512 = "c6425cb53bb7915f5912e49c2b4182fe9875eb60759c15e130ae6d323dc168ad860451
 force-upstream = true
 
 [[package.metadata.build-package.external-files]]
+url = "https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-imex-570-570.148.08-1.x86_64.rpm"
+sha512 = "846e3f2c6dab2bf097771ec01720cf98ccd30f0dcbad09826d708e29f896982642bad866c38fc116c02f6f0a3feea3eaef69034380dc303dbb50aeee079ed01a"
+force-upstream = true
+
+[[package.metadata.build-package.external-files]]
+url = "https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/nvidia-imex-570-570.148.08-1.aarch64.rpm"
+sha512 = "13c1ac2f8650a909ea826fbd41c4ba3378a6c83aa6d38996c2a5509aa49f70a90c023ccaa1e1cad6ecd6808e721cc4effcf02f4cc21f0f2a0fc4043a105ba1bd"
+force-upstream = true
+
+[[package.metadata.build-package.external-files]]
 url = "https://raw.githubusercontent.com/NVIDIA/open-gpu-kernel-modules/570/COPYING"
 sha512 = "f9cee68cbb12095af4b4e92d01c210461789ef41c70b64efefd6719d0b88468b7a67a3629c432d4d9304c730b5d1a942228a5bcc74a03ab1c411c77c758cd938"
 force-upstream = true
--- a/packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
+++ b/packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
@@ -39,6 +39,10 @@ Source3: COPYING
 Source10: https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-fabric-manager-%{tesla_ver}-1.x86_64.rpm
 Source11: https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/nvidia-fabric-manager-%{tesla_ver}-1.aarch64.rpm
 
+# IMEX for GB200
+Source20: https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-imex-%{tesla_ver}-1.x86_64.rpm
+Source21: https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/nvidia-imex-%{tesla_ver}-1.aarch64.rpm
+
 # Common NVIDIA conf files from 200 to 299
 Source200: nvidia-tmpfiles.conf.in
 Source202: nvidia-dependencies-modules-load.conf
@@ -85,8 +89,16 @@ Requires: %{name}-grid
 Summary: NVIDIA fabricmanager config and service files
 Requires: %{name}-tesla(fabricmanager)
 Requires: %{_cross_os}nvlsm
+Requires: %{name}-imex
 
 %description fabricmanager
+%{summary}.
+
+%package imex
+Summary: NVIDIA IMEX config and service files
+Requires: %{name}
+
+%description imex
 %{summary}.
 
 %package open-gpu
@@ -140,6 +152,11 @@ rpm2cpio %{_sourcedir}/nvidia-fabric-manager-%{tesla_ver}-1.%{_cross_arch}.rpm |
 
 # Add the license.
 install -p -m 0644 %{S:2} %{S:3} .
+
+# Extract imex from the rpm via cpio rather than `%%setup` since the
+# correct source is architecture-dependent.
+mkdir imex-%{nvidia_arch}-%{tesla_ver}-archive
+rpm2cpio %{_sourcedir}/nvidia-imex-%{tesla_major}-%{tesla_ver}-1.%{_cross_arch}.rpm | cpio -idmV -D imex-%{nvidia_arch}-%{tesla_ver}-archive
 
 # This recipe was based in the NVIDIA yum/dnf specs:
 # https://github.com/NVIDIA/yum-packaging-precompiled-kmod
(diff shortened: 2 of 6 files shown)


packages/kmod-6.12-nvidia-r570/Cargo.toml
```diff
@@ -36,6 +36,16 @@ url = "https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/nvidi
 sha512 = "c6425cb53bb7915f5912e49c2b4182fe9875eb60759c15e130ae6d323dc168ad860451e2d69932aba8febb350bc7c0ff6d82df625aa0b63d8601e9fa7cc31b2d"
 force-upstream = true
 
+[[package.metadata.build-package.external-files]]
+url = "https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-imex-570-570.148.08-1.x86_64.rpm"
```
arnaldo2792, line 40: I'll create us a task to migrate all of these to:

https://developer.download.nvidia.com/compute/cuda/repos/amzn2023/x86_64/

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -141,6 +153,11 @@ rpm2cpio %{_sourcedir}/nvidia-fabric-manager-%{tesla_ver}-1.%{_cross_arch}.rpm |
 # Add the license.
 install -p -m 0644 %{S:2} %{S:3} .
 
+# Extract fabricmanager from the rpm via cpio rather than `%%setup` since the
```
arnaldo2792: Copy/paste issue: `s/fabricmanager/imex`.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -141,6 +153,11 @@ rpm2cpio %{_sourcedir}/nvidia-fabric-manager-%{tesla_ver}-1.%{_cross_arch}.rpm |
...
+# correct source is architecture-dependent.
+mkdir imex-%{fm_arch}-%{tesla_ver}-archive
```
arnaldo2792: Since `fm_arch` now is used for a different purpose, I think we should rename it to something else. Probably as the first commit in this PR so that the second commit only includes the actual changes of interest.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -708,3 +732,8 @@ popd
 %{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.cfg
 %{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.env
 %{_cross_unitdir}/nvidia-fabricmanager.service
+
+%files imex
+%license imex-%{fm_arch}-%{tesla_ver}-archive/usr/share/licenses/nvidia-imex-%{tesla_major}/LICENSE
```
arnaldo2792: If we are doing this now, we should also update the fabric manager as its RPM also provides a license:

```bash
rpm -q --list nvidia-fabric-manager-570.148.08-1.x86_64.rpm | rg LICENSE
/usr/share/doc/nvidia-fabricmanager/LICENSE
```

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -85,10 +89,18 @@ Requires: %{name}-grid
 Summary: NVIDIA fabricmanager config and service files
 Requires: %{name}-tesla(fabricmanager)
 Requires: %{_cross_os}nvlsm
+Requires: %{name}-imex
```
arnaldo2792, line 92: Are you adding `imex` here because it is a dependency of `fabricmanager` and therefore a dependency of the `tesla` driver?

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
yeazelm: Nice catch, I'll fix.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
yeazelm: I can do that, it felt strange to me as well and we are really stretching it now.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
yeazelm: Correct, Fabric Manager is now responsible for initialization of many of these chips like the B200, and IMEX is only relevant on GB200 at the moment. So the rationale is to keep the dependencies all at the fabric manager level. This is all pretty intertwined at this point but this felt like the right way to have it naturally included.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
bcressey: Is the license different from the overall package license, such that the fm and imex subpackages should have a different license statement?

yeazelm: The Fabric Manager and IMEX licenses are the same. I can drop this, I was just over zealous with tracking it.

yeazelm:
^ Adds in the rename for `fm_arch` to `nvidia_arch` in separate commits along with addressing comments from the PR.

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
```diff
@@ -708,3 +732,8 @@ popd
 %{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.cfg
 %{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.env
 %{_cross_unitdir}/nvidia-fabricmanager.service
+
+%files imex
+
```
bcressey: remove newline after `%files imex`:
```suggestion
```

bcressey (approved):

packages/kmod-6.12-nvidia-r570/kmod-6.12-nvidia-r570.spec
yeazelm: Fixed in recent push.

bcressey (approved):

rpkelly (approved):

packages/kmod-6.12-nvidia-r570/Cargo.toml
arnaldo2792: I'll pause on this ticket for now.

arnaldo2792 (approved):

PR #226, opened 2025-07-29
drop kernel-5.15, R535 NVIDIA kmod packages

### LINKED ISSUE #205: Remove NVIDIA R535 kmod packages from kit
**What I'd like:**
All of the variants in bottlerocket-os/bottlerocket have moved to R570 branch drivers except for `aws-ecs-1-nvidia` and `aws-k8s-1.27-nvidia` which are both going end of life in July [1](https://github.com/bottlerocket-os/bottlerocket/issues/4492) [2](https://github.com/bottlerocket-os/bottlerocket/issues/4580). We should remove `kmod-5.15-nvidia` and `kmod-6.1-nvidia` once those variants are removed so that we can clean up these package which will have no consumers. 

Keep updating these packages even though there are no consumers.

ginglis13:
**Issue number:**

Closes #205

**Description of changes:**

Bottlerocket no longer releases variants built with the kernel-5.15 or
R535 NVIDIA kmod packages. Drop these packages from the kernel-kit

**Testing done:**

```
make ARCH=x86_64 && make ARCH=aarch64
```

**Terms of contribution:**


rpkelly (approved):

yeazelm (approved):
Nice clean up! We should make sure we do a major version bump when we release these changes.

PR #245, opened 2025-08-12
Replace explicit kernel module declarations with glob patterns in kernel-6.12

Tjzeller12:
**Issue number:**
Closes #

**Description of changes:**

Replace 924 explicit kernel module file declarations in kernel-6.12.spec with 10 glob patterns organized by major kernel subsystems (arch, crypto, drivers, fs, kernel, lib, mm, net, security, virt).

**Problem:** The current approach explicitly lists every .ko file, causing kernel update automation to fail when:
- New modules are added to the kernel
- Existing modules are removed 
- Modules are renamed or moved

**Solution:** Use glob patterns like `%{_cross_kmoddir}/kernel/drivers/**/*.%{_ko}` that automatically capture all kernel modules in each subsystem directory, eliminating the need for manual spec file updates.

**Testing done:**

- Confirmed spec file syntax is accepted by the build system without errors
- Verified build process successfully starts and reaches kernel-6.12 compilation phase
- No "file not found" or "invalid glob pattern" errors encountered during build initialization
- Confirmed modules-neuron package uses separate /updates/ directory and won't be affected by kernel/ glob patterns
- Build process was interrupted due to time constraints, but initial validation indicates glob patterns are syntactically correct and resolvable by the build system

**Terms of contribution:**


bcressey:
Listing every module is intentional. It's to ensure that there's an explicit code change whenever modules are added or removed, since removing a module can be a breaking change to downstream consumers.

PR #247, opened 2025-08-18
tools: add makefile targets for kernel configuration generation 

### LINKED ISSUE #174: Modify tools/latest-kernel-full-config.sh to run in the bottlerocket-SDK container

sky1122:
**Issue number:**

Closes #174

**Description of changes:**

Move the work to extract, patch, and merge kernel configurations into a new inner-full-config function, and execute that function in an SDK container in Makefile. As seen in issue https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/173, running this logic on a build machine can break for many reasons. Running inside the SDK world gives the script a stable environment.

**Testing done:**
- [X] generate configs for the kernel 6.1
```
╭─ in ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.1 on develop ✘ (origin/develop)
╰$ diff -u config-full-bottlerocket-x86_64 /home/xxx/workplace/larvacr_branch/bottlerocket-kernel-kit/packages/kernel-6.1/config-full-bottlerocket-x86_64
╭─ in ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.1 on develop ✘ (origin/develop)
╰$ diff -u config-full-bottlerocket-aarch64 /home/xxx/workplace/larvacr_branch/bottlerocket-kernel-kit/packages/kernel-6.1/config-full-bottlerocket-aarch64
╭─ ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.1 on develop ✘ (origin/develop)
╰$
```
diff generate same
- [x] generate config for the kernel 6.12
```
╭─in ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.12 on develop ✘ (origin/develop)
╰$ diff -u config-full-bottlerocket-aarch64 /home/xxx/workplace/larvacr_branch/bottlerocket-kernel-kit/packages/kernel-6.12/config-full-bottlerocket-aarch64
╭─in ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.12 on develop ✘ (origin/develop)
╰$ diff -u config-full-bottlerocket-x86_64 /home/xxx/workplace/larvacr_branch/bottlerocket-kernel-kit/packages/kernel-6.12/config-full-bottlerocket-x86_64
╭─ in ~/workplace/repos/bottlerocket-kernel-kit/packages/kernel-6.12 on develop ✘ (origin/develop)
╰$
```
- [x] change and test the kernel update automation
    - successfully generate the script

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ TOOLS_DIR := $(TOP)tools
 TWOLITER_DIR := $(TOOLS_DIR)/twoliter
 TWOLITER := $(TWOLITER_DIR)/twoliter
 CARGO_HOME := $(TOP).cargo
+KERNEL_CONFIG_SCRIPT := $(TOOLS_DIR)/latest-kernel-full-config.sh
 
 TWOLITER_VERSION ?= "0.11.0"
 TWOLITER_SHA256_AARCH64 ?= "418f762baf1698472af697d40bc4877dd5fa0b033fda31250b3d38e02d72a289"
@@ -20,6 +21,67 @@ endif
 
 
 all: build
+
+merge-kernel-configs:
+	@echo "Validating parameters for merge-kernel-configs..."
+	@usage_msg="Usage: make merge-kernel-configs RPM_FILE=/path/to/kernel-source.rpm SDK_IMAGE=image:tag KVER=x.y"; \
+	validate_param() { \
+		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
+		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
+	}; \
+	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
+	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
+	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
+
+	@tmpdir=$$(mktemp -d); \
+	if [ $$? -ne 0 ]; then \
+		echo "Error: Failed to create temporary directory"; \
+		exit 1; \
+	fi; \
+	cp "$(RPM_FILE)" "$${tmpdir}/kernel-source.rpm"; \
+	if [ $$? -ne 0 ]; then \
+		echo "Error: Failed to copy RPM file to temporary directory"; \
+		rm -rf "$${tmpdir}"; \
+		exit 1; \
+	fi; \
+	cd packages/kernel-$(KVER) && \
+	docker run --rm \
+		-v "$(PWD)/packages/kernel-$(KVER)/":/kernel-package \
+		-v "$${tmpdir}":/work \
+		-v "$(PWD)/packages/microcode/":/microcode \
+		-v "$(PWD)/tools/latest-kernel-full-config.sh":/latest-kernel-full-config.sh \
+		--user "$$(id -u):$$(id -g)" \
+		--name "kernel-$(KVER)-inner-full" \
+		"$(SDK_IMAGE)" \
+		bash -c "source /latest-kernel-full-config.sh && inner_full_config"; \
+	docker_exit_code=$$?; \
+	rm -rf "$${tmpdir}"; \
+	if [ $$docker_exit_code -ne 0 ]; then \
+		echo "Error: Docker container execution failed with exit code $$docker_exit_code"; \
+		exit $$docker_exit_code; \
+	fi; \
+	echo "Successfully merged kernel configs for version $(KVER)"
+
+full-config:
+	@if [ -z "$(RPM_FILE)" ]; then \
+		echo "Error: RPM_FILE parameter is required"; \
+		echo "Usage: make full-config RPM_FILE=/path/to/kernel-source.rpm"; \
+		exit 1; \
+	fi; \
+	echo "Checking kernel configuration dependencies..."; \
+	source $(KERNEL_CONFIG_SCRIPT) && check_dependencies; \
+	echo "All dependencies are available."; \
+	rpm_file=$$(realpath "$(RPM_FILE)"); \
+	if [ ! -f "$${rpm_file}" ]; then \
+		echo "Error: RPM file not found: $${rpm_file}"; \
+		exit 1; \
+	fi; \
+	sdk_image=$$(source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk); \
+	kver=$$(rpm --query --nosignature --queryformat '%{VERSION}' "$${rpm_file}" | sed 's/\.[^.]*$$//'); \
+	echo "Processing kernel version: $${kver}"; \
+	echo "Using SDK image: $${sdk_image}"; \
+	echo "Using RPM file: $${rpm_file}"; \
+	$(MAKE) merge-kernel-configs RPM_FILE="$${rpm_file}" SDK_IMAGE="$${sdk_image}" KVER="$${kver}"
 
 prep:
 	@mkdir -p $(TWOLITER_DIR)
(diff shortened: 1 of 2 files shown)


tools/latest-kernel-full-config.sh
```diff
@@ -2,17 +2,7 @@
```
ginglis13, line 2: Let's update the `usage` to make it clear now that the functions provided by this script are intended to be run inside the bottlerocket-sdk

tools/latest-kernel-full-config.sh
```diff
@@ -70,41 +60,40 @@ resolve_bottlerocket_sdk() {
     echo "${registry}/${name}:v${version}"
 }
 
-# expect $pwd to be packages/kernel-${kver}
-merge_kernel_configs() {
-    rpm_file=$1
-    sdk_image=$2
-    kernel_path=$PWD
+# Function to perform the kernel configuration merging inside Docker container
+# This replaces the functionality of inner-full-config.sh
```
ginglis13: this line of the comment doesn't seem necessary

Makefile
```diff
@@ -21,6 +22,42 @@ endif
...
 
+merge-kernel-configs:
+	@echo "Merging kernel configs for version $(KVER)..."
+	@tmpdir=$$(mktemp -d); \
+	cp "$(RPM_FILE)" "$${tmpdir}/kernel-source.rpm"; \
+	cd packages/kernel-$(KVER) && \
+	docker run --rm \
+		-v "$(PWD)/packages/kernel-$(KVER)/":/kernel-package \
+		-v "$${tmpdir}":/work \
+		-v "$(PWD)/":/package \
```
ginglis13: it looks like we only need `/packages/microcode` from this volume; should we just mount that instead?

sky1122: Yes, the only usage is in here. Change to just mount microcode.
```
microcode_cfg="/package/packages/microcode/${microcode_file}"
```

sky1122:
force pushed to addressed the feedbacks

Makefile
```diff
@@ -21,6 +22,43 @@ endif
 
 all: build
 
+merge-kernel-configs:
+	@echo "Merging kernel configs for version $(KVER)..."
```
ginglis13: should we include some arg checking for `RPM_FILE`, `SDK_IMAGE`, and `KVER` here? if I want to invoke this rule directly without proper inputs, current behavior is:

```
$ make merge-kernel-configs
Merging kernel configs for version ...
cp: cannot stat '': No such file or directory
/bin/sh: line 3: cd: packages/kernel-: No such file or directory
```

sky1122: thank you and good catch! I was assuming this will only be used in the `full-config` and only separated them for the purpose of shorten and keeping function simple. 

Add arg checking.

sky1122:
force pushed to add arg checking in `merge-kernel-configs`

ginglis13 (approved):

dhwaniserai (approved):

sky1122:
force pushed to squash the commit

Makefile
```diff
@@ -21,6 +22,67 @@ endif
...
 
+merge-kernel-configs:
+	@echo "Validating parameters for merge-kernel-configs..."
+	@usage_msg="Usage: make merge-kernel-configs RPM_FILE=/path/to/kernel-source.rpm SDK_IMAGE=image:tag KVER=x.y"; \
+	validate_param() { \
+		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
+		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
+	}; \
+	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
+	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
```
arnaldo2792, line 33: Please use:

```Makefile
@test $${RPM_FILE?not set!}
```
For the other variables as well

Makefile
```diff
@@ -21,6 +22,67 @@ endif
...
+	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
```
arnaldo2792, line 34: Your script is not consistent, you are mixing `@echo` and `echo`.

Makefile
```diff
@@ -21,6 +22,67 @@ endif
...
+
+	@tmpdir=$$(mktemp -d); \
```
arnaldo2792, line 36: You could put this at the beginning of the Makefile and do:

```Makefile
TMP_DIR := (shell mktep -d)
```

Makefile
```diff
@@ -21,6 +22,67 @@ endif
...
+	if [ $$? -ne 0 ]; then \
+		echo "Error: Failed to create temporary directory"; \
+		exit 1; \
+	fi; \
+	cp "$(RPM_FILE)" "$${tmpdir}/kernel-source.rpm"; \
+	if [ $$? -ne 0 ]; then \
+		echo "Error: Failed to copy RPM file to temporary directory"; \
+		rm -rf "$${tmpdir}"; \
+		exit 1; \
+	fi; \
```
arnaldo2792, line 46: No validations needed here, `make` will fail if any of the commands `exit 1`.

Makefile
```diff
@@ -21,6 +22,67 @@ endif
...
+	rpm_file=$$(realpath "$(RPM_FILE)"); \
+	if [ ! -f "$${rpm_file}" ]; then \
+		echo "Error: RPM file not found: $${rpm_file}"; \
+		exit 1; \
+	fi; \
+	sdk_image=$$(source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk); \
+	kver=$$(rpm --query --nosignature --queryformat '%{VERSION}' "$${rpm_file}" | sed 's/\.[^.]*$$//'); \
+	echo "Processing kernel version: $${kver}"; \
+	echo "Using SDK image: $${sdk_image}"; \
+	echo "Using RPM file: $${rpm_file}"; \
```
arnaldo2792, line 83: Instead of inserting a bash script, you should aim to use only `Makefile` idioms, and use scripts only when you can't achieve what you want with just that. All of these can be replaced with `Makefile` idioms.

tools/latest-kernel-full-config.sh
```diff
@@ -114,99 +109,32 @@ merge_kernel_configs() {
...
-            -w /linux-"${version}" \
-            -u "$(id -u)":1000 \
-            --name "${arch}-kernel-${version}-config-merger" \
-            "${sdk_image}" \
-            ./scripts/kconfig/merge_config.sh \
-            "${script_args[@]}"
-
-        mv -f "bottlerocket_${arch}_defconfig" "${kernel_path}/config-full-bottlerocket-${arch}"
-        popd || bail "Could not move around - 'popd' failed. Lets stop before we break anything further."
+        ARCH=${karch} ./scripts/kconfig/merge_config.sh "${script_args[@]}"
```
arnaldo2792, line 136: You can prefix `CROSS_COMPILE` and `KCONFIG_CONFIG` here. instead of `export`ing them.

tools/latest-kernel-full-config.sh
```diff
@@ -26,16 +16,23 @@ usage() {
 Usage: $(basename "$0") -r RPM_FILE
 
 Extract kernel configurations from an RPM, merge with Bottlerocket's, and write out to a file, per architecture (x86_64 and aarch64).
-Run from the top-level bottlerocket-kernel-kit directory with the following parameters:
+
+This script provides functions intended to be run inside the bottlerocket-sdk container.
+For typical usage, run 'make full-config RPM_FILE=/path/to/kernel-source.rpm' from the top-level bottlerocket-kernel-kit directory.
+
+Direct script usage (for development/debugging):
```
arnaldo2792, line 23: I don't agree with the usage of the script out of the `Makefile`, the intention is to use the script with the `Makefile` and use it only with the Bottlerocket SDK. Lets add friction to prevent users from trying to run it outside the Bottlerocket SDK.

PR #258, opened 2025-08-26
Makefile: fix targets `merge-kernel-configs` and `full-config`

### LINKED ISSUE #174: Modify tools/latest-kernel-full-config.sh to run in the bottlerocket-SDK container

sky1122:
**Issue number:**

Closes #174

**Description of changes:**
Refactored Makefile targets merge-kernel-configs and full-config to follow standard Make idioms. Cleaned up latest-kernel-full-config.sh script, corrected environment variables for cross-compilation support and added check for running only in SDK container.

I keep creating the temp directory under the target so that I can clean the temp directory after exit. 

**Testing done:**
Generating the config-full-bottlerocket-aarch64 and config-full-bottlerocket-x86_64 for 6.1 kernel and diff with the one in repo

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,66 +22,8 @@ endif
 
 all: build
 
-merge-kernel-configs:
-	@echo "Validating parameters for merge-kernel-configs..."
-	@usage_msg="Usage: make merge-kernel-configs RPM_FILE=/path/to/kernel-source.rpm SDK_IMAGE=image:tag KVER=x.y"; \
-	validate_param() { \
-		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
-		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
-	}; \
-	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
-	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
-	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
-
-	@tmpdir=$$(mktemp -d); \
-	if [ $$? -ne 0 ]; then \
-		echo "Error: Failed to create temporary directory"; \
-		exit 1; \
-	fi; \
-	cp "$(RPM_FILE)" "$${tmpdir}/kernel-source.rpm"; \
-	if [ $$? -ne 0 ]; then \
-		echo "Error: Failed to copy RPM file to temporary directory"; \
-		rm -rf "$${tmpdir}"; \
-		exit 1; \
-	fi; \
-	cd packages/kernel-$(KVER) && \
-	docker run --rm \
-		-v "$(PWD)/packages/kernel-$(KVER)/":/kernel-package \
-		-v "$${tmpdir}":/work \
-		-v "$(PWD)/packages/microcode/":/microcode \
-		-v "$(PWD)/tools/latest-kernel-full-config.sh":/latest-kernel-full-config.sh \
-		--user "$$(id -u):$$(id -g)" \
-		--name "kernel-$(KVER)-inner-full" \
-		"$(SDK_IMAGE)" \
-		bash -c "source /latest-kernel-full-config.sh && inner_full_config"; \
-	docker_exit_code=$$?; \
-	rm -rf "$${tmpdir}"; \
-	if [ $$docker_exit_code -ne 0 ]; then \
-		echo "Error: Docker container execution failed with exit code $$docker_exit_code"; \
-		exit $$docker_exit_code; \
-	fi; \
-	echo "Successfully merged kernel configs for version $(KVER)"
-
 full-config:
-	@if [ -z "$(RPM_FILE)" ]; then \
-		echo "Error: RPM_FILE parameter is required"; \
-		echo "Usage: make full-config RPM_FILE=/path/to/kernel-source.rpm"; \
-		exit 1; \
-	fi; \
-	echo "Checking kernel configuration dependencies..."; \
-	source $(KERNEL_CONFIG_SCRIPT) && check_dependencies; \
-	echo "All dependencies are available."; \
-	rpm_file=$$(realpath "$(RPM_FILE)"); \
-	if [ ! -f "$${rpm_file}" ]; then \
-		echo "Error: RPM file not found: $${rpm_file}"; \
-		exit 1; \
-	fi; \
-	sdk_image=$$(source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk); \
-	kver=$$(rpm --query --nosignature --queryformat '%{VERSION}' "$${rpm_file}" | sed 's/\.[^.]*$$//'); \
-	echo "Processing kernel version: $${kver}"; \
-	echo "Using SDK image: $${sdk_image}"; \
-	echo "Using RPM file: $${rpm_file}"; \
-	$(MAKE) merge-kernel-configs RPM_FILE="$${rpm_file}" SDK_IMAGE="$${sdk_image}" KVER="$${kver}"
+	./tools/docker-run.sh "/bottlerocket-kernel-kit/tools/latest-kernel-full-config.sh"
 
 prep:
 	@mkdir -p $(TWOLITER_DIR)
@@ -119,4 +61,4 @@ endif
 twoliter: prep
 	@$(TWOLITER_MAKE) $(TWOLITER_MAKE_ARGS)
 
-.PHONY: prep update fetch build publish twoliter merge-kernel-configs full-config
+.PHONY: prep update fetch build publish twoliter full-config
--- a/tools/docker-run.sh
+++ b/tools/docker-run.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+set -e -o pipefail
+
+
+bail() {
+    if [[ $# -gt 0 ]]; then
+        >&2 echo "Error: $*"
+    fi
+    exit 1
+}
+
+SCRIPT_PATH="$1"
+
+source="$(tq -r ".sdk.source" -f Twoliter.lock)"
+version="$(tq -r ".sdk.version" -f Twoliter.lock)"
+_name="$(tq -r ".sdk.name" -f Twoliter.lock)"
+_registry="${source%/*}"
+
+registry="$(tq -r ".bottlerocket.bottlerocket-sdk.registry" -f Twoliter.override 2>/dev/null || echo "$_registry")"
+name="$(tq -r ".bottlerocket.bottlerocket-sdk.name" -f Twoliter.override 2>/dev/null || echo "$_name")"
+
+SDK="${registry}/${name}:v${version}"
+
+docker run --rm \
+    -v "$(pwd):/bottlerocket-kernel-kit" \
+    --user "$(id -u):$(id -g)" \
+    "${SDK}" \
+    bash "${SCRIPT_PATH}"
+
--- a/tools/latest-kernel-full-config.sh
+++ b/tools/latest-kernel-full-config.sh
@@ -2,6 +2,35 @@
 
 set -e -o pipefail
 
+KERNEL_KIT_DIR="/bottlerocket-kernel-kit"
+MICROCODE_DIR="${KERNEL_KIT_DIR}/packages/microcode"
+
+# Usage information
+usage() {
+    cat << EOF
+Usage: $0
+
+This script generates and validates full kernel configurations for all available
+kernel versions in the Bottlerocket kernel kit.
+
+IMPORTANT: This script is designed to run inside the Bottlerocket SDK container
+and should typically be invoked via the Makefile target:
+
+    make full-config
+
+Running this script directly outside the SDK container will fail because it
+requires the proper build environment and mounted paths.
+
+The script will:
+1. Discover all kernel-* packages with available RPMs
+2. Extract and patch kernel sources
+3. Merge Bottlerocket-specific configurations
+4. Generate full configuration files
+5. Validate that all required configs are present
+
+EOF
+}
+
 # Common error handling
 bail() {
     if [[ $# -gt 0 ]]; then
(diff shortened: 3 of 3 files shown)


sky1122:
force pushed to rebase

Makefile
```diff
@@ -23,65 +57,40 @@ endif
...
-	rpm_file=$$(realpath "$(RPM_FILE)"); \
-	if [ ! -f "$${rpm_file}" ]; then \
-		echo "Error: RPM file not found: $${rpm_file}"; \
-		exit 1; \
-	fi; \
-	sdk_image=$$(source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk); \
+	@test $${RPM_FILE?not set!}
+	@echo "Checking kernel configuration dependencies..."
+	$(eval MISSING_TOOLS := $(shell for tool in rpm2cpio cpio tar docker tq; do command -v $$tool >/dev/null 2>&1 || echo $$tool; done))
+	$(if $(MISSING_TOOLS),$(error DEPENDENCY ERROR: Missing tools: $(MISSING_TOOLS). Please install them in your PATH.))
```
arnaldo2792: This could be a target, which is a dependency of this target.

Makefile
```diff
@@ -23,65 +57,40 @@ endif
...
-	echo "All dependencies are available."; \
-	rpm_file=$$(realpath "$(RPM_FILE)"); \
-	if [ ! -f "$${rpm_file}" ]; then \
-		echo "Error: RPM file not found: $${rpm_file}"; \
-		exit 1; \
-	fi; \
-	sdk_image=$$(source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk); \
+	@test $${RPM_FILE?not set!}
+	@echo "Checking kernel configuration dependencies..."
+	$(eval MISSING_TOOLS := $(shell for tool in rpm2cpio cpio tar docker tq; do command -v $$tool >/dev/null 2>&1 || echo $$tool; done))
```
arnaldo2792: You are doing your validation on at least `tq`, after you have already attempted to use it in the first section of your Makefile.

Makefile
```diff
@@ -5,6 +5,40 @@ TWOLITER := $(TWOLITER_DIR)/twoliter
...
+SDK_NAME_DEFAULT := $(shell tq -r ".sdk.name" -f Twoliter.lock 2>/dev/null)
+SDK_REGISTRY_DEFAULT := $(shell echo "$(SDK_SOURCE)" | sed 's|/[^/]*$$||')
+
+# Check for overrides
+SDK_REGISTRY_OVERRIDE := $(shell tq -r ".bottlerocket.bottlerocket-sdk.registry" -f Twoliter.override 2>/dev/null)
+SDK_NAME_OVERRIDE := $(shell tq -r ".bottlerocket.bottlerocket-sdk.name" -f Twoliter.override 2>/dev/null)
+
+# Use override values if available, otherwise use defaults
+SDK_REGISTRY := $(if $(SDK_REGISTRY_OVERRIDE),$(SDK_REGISTRY_OVERRIDE),$(SDK_REGISTRY_DEFAULT))
+SDK_NAME := $(if $(SDK_NAME_OVERRIDE),$(SDK_NAME_OVERRIDE),$(SDK_NAME_DEFAULT))
```
arnaldo2792: Don't you have to set this after `SDK_REGISTRY_DEFAULT`?

Makefile
```diff
@@ -23,65 +57,40 @@ endif
...
-	@usage_msg="Usage: make merge-kernel-configs RPM_FILE=/path/to/kernel-source.rpm SDK_IMAGE=image:tag KVER=x.y"; \
-	validate_param() { \
-		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
-		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
-	}; \
-	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
-	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
-	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
-
+	@test $${SDK_IMAGE?not set!}
```
arnaldo2792: Shouldn't this be the `RESOLVED_SDK_IMAGE`?

Makefile
```diff
@@ -23,65 +57,40 @@ endif
...
+	@echo "All dependencies are available."
+	@echo "Processing RPM file and extracting metadata..."
+	@rpm_file=$$(realpath "$(RPM_FILE)"); \
+	test -f "$${rpm_file}"; \
 	kver=$$(rpm --query --nosignature --queryformat '%{VERSION}' "$${rpm_file}" | sed 's/\.[^.]*$$//'); \
-	echo "Processing kernel version: $${kver}"; \
-	echo "Using SDK image: $${sdk_image}"; \
-	echo "Using RPM file: $${rpm_file}"; \
-	$(MAKE) merge-kernel-configs RPM_FILE="$${rpm_file}" SDK_IMAGE="$${sdk_image}" KVER="$${kver}"
+	$(MAKE) merge-kernel-configs RPM_FILE="$${rpm_file}" SDK_IMAGE="$(RESOLVED_SDK_IMAGE)" KVER="$${kver}"
```
arnaldo2792: This is weird, you are calling a make target from within a make target. You should do this instead:

```Makefile
# Call merge-kernel-configs before full-config is called
full-config: merge-kernel-configs
    # Do the full-config stuff here
```

Makefile
sky1122: My purpose is using `full-config` to pass this `SDK_IMAGE` to `merge-kernel-configs` . That's why I didn't directly use `RESOLVED_SDK_IMAGE `
```
$(MAKE) merge-kernel-configs RPM_FILE="$${rpm_file}" SDK_IMAGE="$(RESOLVED_SDK_IMAGE)" KVER="$${kver}"
```

Let me check how to make this better by following the next comment

Makefile
sky1122: putting in after checking [Twoliter.override](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/pull/258/files/dcb742307edb88a777aa4bdb0510a287dc14e3ee#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R15-R16) to make sure if there is a Twoliter.override so that we get the sdk value from that. Similar with the [original script](https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/develop/tools/latest-kernel-full-config.sh#L54C1-L54C25)

sky1122: I have some repeat code probably from rebase. delete them....

Makefile
sky1122: I was thinking this can be separated logic for the `full-config`. But after revision, I think I can combine these two in one function based on the fact `full-config` was doing dependency check and getting the parameters that `merge-kernel-configs` needed.

sky1122:
force push to address feedback. Main change
- Moving resolve sdk image logic back to `latest-kernel-full-config.sh` since IMO it looks more clear for the Makefile
- move the dependency check to a target
- merge `full-config` and `merge-kernel-configs` to one target.

sky1122:
force pushed to fix the commit message

rpkelly (approved):

sky1122:
force push to rebase

Makefile
```diff
@@ -22,66 +22,38 @@ endif
...
-		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
-		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
-	}; \
-	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
-	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
-	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
-
+check-dependencies:
+	@echo "Checking kernel configuration dependencies..."
+	$(eval MISSING_TOOLS := $(shell for tool in rpm2cpio cpio tar docker tq; do command -v $$tool >/dev/null 2>&1 || echo $$tool; done))
```
ginglis13: now that we've containerized things, aren't we getting `rpm2cpio` and `cpio` from the SDK image?

sky1122: yes. `tq` and `docker` need to have. will change that

sky1122:
force pushed to change the dependency check context.
force pushed to make the commits have correct content.

Makefile
```diff
@@ -22,66 +22,38 @@ endif
...
-	@usage_msg="Usage: make merge-kernel-configs RPM_FILE=/path/to/kernel-source.rpm SDK_IMAGE=image:tag KVER=x.y"; \
-	validate_param() { \
-		param_name="$${1}"; param_value="$${2}"; error_msg="$${3}"; \
-		[ -n "$${param_value}" ] || { echo "Error: $${error_msg}"; echo "$${usage_msg}"; exit 1; }; \
-	}; \
-	validate_param "SDK_IMAGE" "$(SDK_IMAGE)" "SDK_IMAGE parameter is required"; \
-	validate_param "KVER" "$(KVER)" "KVER parameter is required"; \
-	echo "All parameter checks passed. Merging kernel configs for version $(KVER)..."
-
+check-dependencies:
```
arnaldo2792: I talked with @sky1122 offline. The idea is to make this happen:

```Makefile
full-config:
    docker run <whatever> full-config
```

So that all the setup happens in the container, and all the evaluations/validations occur in the script that is executed in the container.

sky1122:
Here is the reason why I still keep these shell command in the Makefile:
- need `RPM_PATH` to get the kernel-xx-rpm package
- SDK_IMAGE: need to get to pass to docker as the image name
- KVER: need to auto resolve from the rpm package. Only run `full-config` when a related kernel rpm provided. not always running all
- KERNEL_PACKAGE_DIR: need to move the final full-config outside the directory

Makefile
```diff
@@ -22,66 +22,38 @@ endif
...
+#    Need to get the kernel version based on the input RPM
+prepare-kernel-config:
+	@echo "Checking kernel configuration dependencies..."
+	@test $${RPM_FILE?not set!}
+	$(eval MISSING_TOOLS := $(shell for tool in docker tq; do command -v $$tool >/dev/null 2>&1 || echo $$tool; done))
+	$(if $(MISSING_TOOLS),$(error DEPENDENCY ERROR: Missing tools: $(MISSING_TOOLS). Please install them in your PATH.))
+	$(eval export RPM_PATH := $(shell realpath "$(RPM_FILE)"))
+	$(eval export SDK_IMAGE := $(shell cd $(TOP) && source $(KERNEL_CONFIG_SCRIPT) && resolve_bottlerocket_sdk))
+	$(eval export KVER := $(shell rpm --query --nosignature --queryformat '%{VERSION}' "$(RPM_FILE)" | sed 's/\.[^.]*$$//'))
+	$(eval export KERNEL_PACKAGE_DIR := $(TOP)packages/kernel-$(KVER))
```
arnaldo2792: You can do all of this within the container, if you mount `$(TOP)/packages/`

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +97,91 @@ validate_kernel_configs() {
...
+    # Ensure this script is run within the Bottlerocket SDK container by checking mounted directories
+    if [[ ! -d "/kernel-package" ]] || [[ ! -f "/rpm-file/kernel-source.rpm" ]] || [[ ! -d "/microcode" ]]; then
+        bail "This script must be run within the Bottlerocket SDK container." \
+             "Expected mount points /kernel-package, /rpm-file/kernel-source.rpm, and /microcode not found." \
+             "Usage: make full-config RPM_FILE=/path/to/kernel-source.rpm"
+    fi
+
+# Create temporary working directory
+tmpdir=$(mktemp -d)
+trap 'rm -rf ${tmpdir}' EXIT
```
arnaldo2792: This line should be on top of the file, instead of here, and should call a cleanup function rather than the once command. But since you are running this in a container, why do you care about the clean up? The filesystem of the container is ephemeral so you don't really have to clean up anything.

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +97,91 @@ validate_kernel_configs() {
...
+# Patches from bottlerocket
+for patch in "${br_patches[@]}"; do
+    echo "Applying bottlerocket patch ${patch}"
+    patch -p1 <"$patch"
+done
+
+popd || bail "Could not move around - 'popd' back to /work failed. Lets stop before we break anything further."
+
+for arch in "x86_64" "aarch64"; do
+
```
arnaldo2792, line 247: The extra space isn't really needed.

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +97,91 @@ validate_kernel_configs() {
...
+    if [ "${arch}" = "aarch64" ]; then
+        karch="arm64"
+        script_args=("../config-${arch}" "${br_cfg}" "${br_cfg_arch}")
+    elif [ "${arch}" = "x86_64" ]; then
+        karch="x86"
+        script_args=("../config-${arch}" "${microcode_cfg}" "${br_cfg}" "${br_cfg_arch}")
+    fi
+
+    ARCH=${karch} CROSS_COMPILE=/usr/bin/${arch}-bottlerocket-linux-gnu- KCONFIG_CONFIG=bottlerocket_${arch}_defconfig ./scripts/kconfig/merge_config.sh "${script_args[@]}"
```
arnaldo2792: You could break down this line to make the script more readable (try to keep your line length under 85 characters)

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +97,91 @@ validate_kernel_configs() {
         bail "Kernel configuration validation failed"
     fi
 }
+
+# Only check container environment if we're running the main script (not just sourcing functions)
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
arnaldo2792: I think you can skip the check to validate if the script is running inside a container. Instead you can perform all the validations that you require for the script to work properly, like you did underneath.

sky1122:
forced push to fix the commit history and remove mount rpm file into SDK container

sky1122:
force pushed to allow when only one kernel has rpm package

tools/docker-run.sh
```diff
@@ -0,0 +1,32 @@
...
+registry="$(tq -r ".bottlerocket.bottlerocket-sdk.registry" -f Twoliter.override 2>/dev/null || echo "$_registry")"
+name="$(tq -r ".bottlerocket.bottlerocket-sdk.name" -f Twoliter.override 2>/dev/null || echo "$_name")"
+
+SDK="${registry}/${name}:v${version}"
+
+# Run the script in docker
+docker run --rm \
+    -v "$(pwd):/bottlerocket-kernel-kit" \
+    --user "$(id -u):$(id -g)" \
+    --name "kernel-full-config" \
```
arnaldo2792: You don't have to give this a name, if anything, it shouldn't be always `kernel-full-config` as the purpose of the script is to be used to run any script within a docker container

tools/docker-run.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
arnaldo2792, line 1: nit: in your commit message, you could provide a bit more details on this new script

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +160,56 @@ validate_kernel_configs() {
...
+KERNEL_KIT_DIR="/bottlerocket-kernel-kit"
+# Ensure this script is run within the Bottlerocket SDK container
+if [[ ! -d "${KERNEL_KIT_DIR}" ]]; then
+    bail "This script must be run within the Bottlerocket SDK container." \
+         "Expected mount point ${KERNEL_KIT_DIR} not found."
+fi
+
+# Check which kernels have RPMs available and process only those
+kernels_to_process=()
+for kernel_pkg in "kernel-6.1" "kernel-6.12"; do
```
arnaldo2792: What if instead of hardcoding the list of kernels, you just generate the configs for all the kernels that you found in the packages directory? That way, we don't have to keep updating the script as we add more kernels.

tools/latest-kernel-full-config.sh
```diff
@@ -10,82 +10,31 @@ bail() {
     exit 1
 }
 
-# Function to display usage information
-usage() {
```
arnaldo2792, line 14: IMO, you still need the `usage` function, probably a bit simplified, at least to explain people that mistakenly ran the script `latest-kernel-full-config.sh` without the `Makefile` target.

tools/latest-kernel-full-config.sh
```diff
@@ -10,82 +10,31 @@ bail() {
...
-
-    # Form the final SDK
-    echo "${registry}/${name}:v${version}"
-}
-
-# Function to perform the kernel configuration merging inside Docker container
-inner_full_config() {
-    pushd /work || bail "Unable to enter /work (bind mount to tmp dir)"
+    local majorminor=${version%.*} # Trim after last '.', e.g. 6.1.132 -> 6.1
+    local KERNEL_PACKAGE_DIR="/bottlerocket-kernel-kit/packages/kernel-${majorminor}"
```
arnaldo2792: capitalized variables are meant to be used for constants, not local variables.

tools/latest-kernel-full-config.sh
```diff
@@ -10,82 +10,31 @@ bail() {
...
-Note: The inner_full_config() function is designed to run within the bottlerocket-sdk container environment
-and expects specific mount points and toolchain paths to be available.
-EOF
-}
-
-usage_error() {
-    >&2 usage
-    bail "$1"
-}
+MICROCODE_DIR="/bottlerocket-kernel-kit/packages/microcode"
```
arnaldo2792: Constants go on the top of the file

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +160,56 @@ validate_kernel_configs() {
         bail "Kernel configuration validation failed"
     fi
 }
+
+# Set paths
+KERNEL_KIT_DIR="/bottlerocket-kernel-kit"
```
arnaldo2792: Same as elsewhere, constants on the top

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +184,79 @@ validate_kernel_configs() {
...
+    shift
+done
+
+# Ensure this script is run within the Bottlerocket SDK container
+if [[ ! -d "${KERNEL_KIT_DIR}" ]]; then
+    echo "❌ This script must be run within the Bottlerocket SDK container."
+    echo "   Expected mount point ${KERNEL_KIT_DIR} not found."
+    echo ""
+    echo "   Please use: make full-config"
+    echo ""
```
arnaldo2792: You should just use `usage` here.

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +184,79 @@ validate_kernel_configs() {
...
+    else
+        bail "No Kernel directory found for ${kernel_dir}"
+    fi
+done
+
+if [[ ${#kernels_to_process[@]} -eq 0 ]]; then
+    bail "No kernel RPMs found in any package directory"
+fi
+
+# Process available kernels
```
arnaldo2792, line 236: Comments should explain why you did what you did, not what's happening. The code is explicit enough to understand what's happening. Same for all other comments.

tools/latest-kernel-full-config.sh
```diff
@@ -2,90 +2,66 @@
...
 
-# Common error handling
-bail() {
-    if [[ $# -gt 0 ]]; then
-        >&2 echo "Error: $*"
-    fi
-    exit 1
-}
+MICROCODE_DIR="/bottlerocket-kernel-kit/packages/microcode"
+KERNEL_KIT_DIR="/bottlerocket-kernel-kit"
```
arnaldo2792: nit:

```suggestion
KERNEL_KIT_DIR="/bottlerocket-kernel-kit"
MICROCODE_DIR="${KERNEL_KIT_DIR}/packages/microcode"
```

tools/latest-kernel-full-config.sh
```diff
@@ -2,90 +2,66 @@
...
+# Function to merge kernel configurations for a specific kernel version
+merge_kernel_configs() {
+    local version="$1"
+    local majorminor="$2"
+    local tmpdir="$3"
 
-# Function to perform the kernel configuration merging inside Docker container
-inner_full_config() {
-    pushd /work || bail "Unable to enter /work (bind mount to tmp dir)"
+    local kernel_package_dir="/bottlerocket-kernel-kit/packages/kernel-${majorminor}"
```
arnaldo2792: You could use `${KERNEL_KIT_DIR}` here

arnaldo2792 (approved):

arnaldo2792 (approved):

tools/latest-kernel-full-config.sh
```diff
@@ -2,90 +2,66 @@
...
-    - tq
+Running this script directly outside the SDK container will fail because it
+requires the proper build environment and mounted paths.
 
-Note: The inner_full_config() function is designed to run within the bottlerocket-sdk container environment
-and expects specific mount points and toolchain paths to be available.
-EOF
-}
+The script will:
+1. Discover all kernel-* packages with available RPMs
```
ginglis13, line 25: What is the behavior when there are multiple RPMs in a kernel package subdir? This PR would force me to have some better hygiene, but currently my kernel-6.12 dir has many RPMs across a few different builds of the kernel-kit:

```
╭─fedora in ~/kernel-kit/packages/kernel-6.12  ‹develop*›
╰─➤  ls *.rpm
aws-neuronx-dkms-2.20.28.0.noarch.rpm  kernel6.12-6.12.25-32.101.amzn2023.src.rpm  kernel6.12-6.12.37-61.105.amzn2023.src.rpm  kernel6.12-6.12.40-64.114.amzn2023.src.rpm
aws-neuronx-dkms-2.21.37.0.noarch.rpm  kernel6.12-6.12.29-33.102.amzn2023.src.rpm  kernel6.12-6.12.40-63.114.amzn2023.src.rpm
```

tools/latest-kernel-full-config.sh
```diff
@@ -213,3 +184,71 @@ validate_kernel_configs() {
...
+fi
+
+# In kernel kit automation, CodePipeline updates each kernel separately.
+# This script processes only kernels that have available RPM files,
+# skipping any kernel packages without corresponding source RPMs.
+kernels_to_process=()
+for kernel_dir in "${KERNEL_KIT_DIR}/packages"/kernel-*; do
+    if [[ -d "${kernel_dir}" ]]; then
+        kernel_pkg=$(basename "${kernel_dir}")
+        rpm_file=$(find "${kernel_dir}" -name "kernel*.src.rpm" | head -1)
```
ginglis13: This should get the latest RPM if you use `tail`:


```suggestion
        rpm_file=$(find "${kernel_dir}" -name "kernel*.src.rpm" | tail -1)
```

ginglis13 (approved):

PR #289, opened 2025-10-13
Full config containerized

arnaldo2792:
**Description of changes:**

The last revision of the script introduced a dependency on `tq` to retrieve which SDK version should be used, but it wasn't clear which `tq` (`tq` the rust binary or `tq` the go binary).

Drop this hardcoded dependency on the running host by calling the [`tq` container] preventing the container from making any network calls. Additionally, allow users provide their own SDK url as they need to instead of extracting the overridden SDK from a `Twoliter.override`.

The last change in the series retrieves the current kernel sources, so that users don't have to do a build to generate the new kernel configs.

[`tq` container]: https://github.com/mdm-code/tq
**Testing done:**

I ran `make full-config` without `tq` in my system, and without prior builds. I confirmed:
- The script finds the correct SDK version
- The script uses the SDK override when provided
- The script fetches the kernel RPMs in brand new kernel kit tree
- The script generates the config

**Terms of contribution:**

### DIFF
--- a/tools/docker-run.sh
+++ b/tools/docker-run.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -e -o pipefail
 
+TQ_IMAGE=ghcr.io/mdm-code/tq:v2.4.0
 
 bail() {
     if [[ $# -gt 0 ]]; then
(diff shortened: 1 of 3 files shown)


tools/docker-run.sh
```diff
@@ -1,25 +1,30 @@
...
 }
 
-SCRIPT_PATH="$1"
+find_sdk() {
+  docker run --rm \
+      -v "$(pwd):/bottlerocket-kernel-kit:ro" \
+      --user "$(id -u):$(id -g)" \
+      --network none \
+      ghcr.io/mdm-code/tq:v2.4.0 \
+      tq -q "[sdk][source]" /bottlerocket-kernel-kit/Twoliter.lock
```
ginglis13, line 19: 💯  previously we were using https://github.com/mdm-code/tq, which doesn't have this syntax nicety

However, this change has a regression in that it doesn't handle `Twoliter.override` usage:

```
╰─➤  cat Twoliter.override
[bottlerocket.bottlerocket-sdk]
registry = "<redacted>.dkr.ecr.us-west-2.amazonaws.com"
name = "thar-be-a-beta-sdk"

╰─➤  ./tools/twoliter/twoliter update

╰─➤  cat Twoliter.lock
schema-version = 2
kit = []
project-vendor = "Bottlerocket"

[sdk]
name = "bottlerocket-sdk"
version = "0.64.0"
vendor = "bottlerocket"
source = "public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.64.0"
digest = "..."

╰─➤  docker run --rm \
      -v "$(pwd):/bottlerocket-kernel-kit:ro" \
      --user "$(id -u):$(id -g)" \
      --network none \
      ghcr.io/mdm-code/tq:v2.4.0 \
      tq -q "[sdk][source]" /bottlerocket-kernel-kit/Twoliter.lock
public.ecr.aws/bottlerocket/bottlerocket-sdk:v0.64.0
```

This should be printing 

```
<redacted>.dkr.ecr.us-west-2.amazonaws.com/thar-be-a-beta-sdk:v0.64.0
```

given the provided override

ginglis13: I see this was done intentionally - LGTM 😄

ginglis13 (approved):

yeazelm (approved):

sky1122 (approved):

arnaldo2792:
Forced push addresses offline feedback:
- Set TQ_IMAGE as variable on top
- Favor `[[ ]]` instead of `[ ]`

jpculp (approved):

jpculp (approved):

ginglis13 (approved):

PR #294, opened 2025-10-14
Add dependency on the GRID license for NVIDIA k8s device plugin

yeazelm:
**Description of changes:**
The GRID driver requires a license to fully function. This currently is fetched as a best effort but a node can come up without a "Licensed" driver and fail later after some amount of time. This change a new file to confirm the license has been correctly acquired or is not needed. This includes a drop in for nvidia-k8s-device-plugin.service to depend on this file:
```
ExecStartPre=/usr/bin/test -f /etc/drivers/.grid-licensed
```
This will be before the `ExecStart` so that the device plugin will not register GPUs unless the license is not required or properly configured.

**Testing done:**
WIP - I'm updating the testing with the new approach for every use case but the primary use cases (GRID, open-gpu, and proprietary) all provide the correct experience. 

Built this change and confirmed that it does not run on a G6 (which uses open-gpu, not GRID so this isn't required), fails to have the node become ready when `nvidia-gridd` is misconfigured to not start. And the nodes become ready as expected when `nvidia-gridd` starts running normally.

Normal g6f.xlarge
```
bash-5.1# journalctl -u grid-license-check.service
Dec 03 16:50:39 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 16:50:39 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 16:50:39 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 16:50:39 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 16:50:40 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 16:50:41 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: grid-license-check.service: Deactivated successfully.
Dec 03 16:50:41 ip-192-168-52-46.us-west-2.compute.internal systemd[1]: Finished GRID License Check.
bash-5.1# ls -al /etc/drivers/.grid-licensed
-rw-r--r--. 1 root root 0 Dec  3 16:50 /etc/drivers/.grid-licensedt-2.compute.internal systemd[1]: Finished GRID License Check.
```

Broken nvidia-gridd results in the node becoming ready but in degraded systemd state and no GPUs offered:
```
bash-5.1# systemctl status
● ip-192-168-71-6.us-west-2.compute.internal
    State: degraded
    Units: 461 loaded (incl. loaded aliases)
     Jobs: 3 queued
   Failed: 1 units
    Since: Wed 2025-12-03 21:24:43 UTC; 46min ago
  systemd: 257.9
  Tainted: unmerged-bin
   CGroup: /


bash-5.1# journalctl -u grid-license-check.service
Dec 03 21:24:55 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 21:24:55 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 21:24:55 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 21:24:55 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 21:24:56 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 21:24:56 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 21:24:56 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 21:24:56 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 21:24:57 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 21:24:57 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 21:24:57 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 21:24:57 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 21:24:59 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 21:25:00 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 21:25:00 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 21:25:00 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 21:25:03 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 21:25:03 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Main process exited, code=exited, status=1/FAILURE
Dec 03 21:25:03 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
Dec 03 21:25:03 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: Failed to start GRID License Check.
Dec 03 21:25:06 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Start request repeated too quickly.
Dec 03 21:25:06 ip-192-168-71-6.us-west-2.compute.internal systemd[1]: grid-license-check.service: Failed with result 'exit-code'.
.... 
# this continues as long as the instances is up but no license exists
```

Output from journal on a g6.2xlarge which uses the fallback:
```
bash-5.1# journalctl -u grid-license-check.service
Dec 03 16:52:46 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 16:52:46 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: grid-license-check.service: Skipped due to 'exec-condition'.
Dec 03 16:52:46 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Condition check resulted in GRID License Check being skipped.
Dec 03 16:52:48 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Starting GRID License Check...
Dec 03 16:52:48 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: grid-license-check.service: Skipped due to 'exec-condition'.
Dec 03 16:52:48 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Condition check resulted in GRID License Check being skipped.
bash-5.1# journalctl -u open-gpu-license-fallback.service
Dec 03 16:52:47 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Starting Open GPU GRID License Check Fallback...
Dec 03 16:52:48 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Finished Open GPU GRID License Check Fallback.
bash-5.1# journalctl -u nvidia-k8s-device-plugin
Dec 03 16:52:48 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Starting Start NVIDIA kubernetes device plugin...
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal systemd[1]: Started Start NVIDIA kubernetes device plugin.
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.175860    3114 main.go:235] "Starting NVIDIA Device Plugin" version="unknown"
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.175889    3114 main.go:238] Starting FS watcher for /var/lib/kubelet/device-plugins
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.175949    3114 main.go:245] Starting OS watcher.
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.176276    3114 main.go:260] Starting Plugins.
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.176288    3114 main.go:317] Loading configuration.
Dec 03 16:52:49 ip-192-168-84-108.us-west-2.compute.internal nvidia-device-plugin[3114]: I1203 16:52:49.177360    3114 main.go:342] Updating config with default resource matching patterns.
....

# The file exists to let the device plugin start
bash-5.1# ls -al /etc/drivers/.grid-licensed
-rw-r--r--. 1 root root 0 Dec  3 16:52 /etc/drivers/.grid-licensed
```

**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.1-nvidia-r570/grid-license-check.service
+++ b/packages/kmod-6.1-nvidia-r570/grid-license-check.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=GRID License Check
+RefuseManualStart=true
+RefuseManualStop=true
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+ExecStart=-/usr/bin/truncate /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/nvidia-smi -q
+ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+StandardOutput=append:/tmp/.nvidia-gridd-license
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r570/grid-license-check.timer
+++ b/packages/kmod-6.1-nvidia-r570/grid-license-check.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=GRID License Check Timer
+RefuseManualStart=true
+
+[Timer]
+Unit=grid-license-check.service
+OnBootSec=5s
+OnUnitActiveSec=2s
+AccuracySec=1s
+
+[Install]
+WantedBy=timers.target
--- a/packages/kmod-6.1-nvidia-r570/grid-license-file-check.conf
+++ b/packages/kmod-6.1-nvidia-r570/grid-license-file-check.conf
@@ -1,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/bin/test -f /etc/drivers/.grid-licensed
--- a/packages/kmod-6.1-nvidia-r570/tesla-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r570/tesla-license-fallback.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver tesla
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+RemainAfterExit=true
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r580/tesla-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r580/tesla-license-fallback.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver tesla
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+RemainAfterExit=true
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.12-nvidia-r570/grid-license-check.service
+++ b/packages/kmod-6.12-nvidia-r570/grid-license-check.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=GRID License Check
+RefuseManualStart=true
+RefuseManualStop=true
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+ExecStart=-/usr/bin/truncate /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/nvidia-smi -q
+ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+StandardOutput=append:/tmp/.nvidia-gridd-license
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.12-nvidia-r580/tesla-license-fallback.service
+++ b/packages/kmod-6.12-nvidia-r580/tesla-license-fallback.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver tesla
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+RemainAfterExit=true
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r570/kmod-6.1-nvidia-r570.spec
+++ b/packages/kmod-6.1-nvidia-r570/kmod-6.1-nvidia-r570.spec
@@ -49,6 +49,11 @@ Source206: nvidia-persistenced.service
 Source207: fabricmanager.env
 Source208: gridd.conf
 Source209: nvidia-gridd.service
+Source210: grid-license-check.service
+Source211: grid-license-check.timer
+Source212: open-gpu-license-fallback.service
+Source213: tesla-license-fallback.service
+Source214: grid-license-file-check.conf
 
 # NVIDIA tesla conf files from 300 to 399
 Source300: nvidia-tesla-tmpfiles.conf
@@ -394,7 +399,9 @@ install kernel-open/nvidia-drm.ko %{buildroot}%{_cross_datadir}/nvidia/grid/driv
 # Install nvidia-gridd and related files
 install -m 755 nvidia-gridd %{buildroot}%{_cross_bindir}/nvidia-gridd
 install -m 644 %{S:208} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia/gridd.conf
-install -p -m 0644 %{S:209} %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:209} %{S:210} %{S:211} %{S:212} %{S:213} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_unitdir}/nvidia-k8s-device-plugin.service.d
+install -p -m 0644 %{S:214} %{buildroot}%{_cross_unitdir}/nvidia-k8s-device-plugin.service.d
 popd
 # End GRID driver
 %endif
@@ -722,6 +729,11 @@ popd
 %{_cross_bindir}/nvidia-gridd
 %{_cross_factorydir}%{_cross_sysconfdir}/nvidia/gridd.conf
 %{_cross_unitdir}/nvidia-gridd.service
+%{_cross_unitdir}/grid-license-check.service
+%{_cross_unitdir}/grid-license-check.timer
+%{_cross_unitdir}/open-gpu-license-fallback.service
+%{_cross_unitdir}/tesla-license-fallback.service
+%{_cross_unitdir}/nvidia-k8s-device-plugin.service.d/grid-license-file-check.conf
 
 %{_cross_datadir}/nvidia/grid/drivers/nvidia.ko
 %{_cross_datadir}/nvidia/grid/drivers/nvidia-uvm.ko
--- a/packages/kmod-6.1-nvidia-r570/open-gpu-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r570/open-gpu-license-fallback.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Open GPU GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver open-gpu
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+RemainAfterExit=true
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r580/grid-license-check.service
+++ b/packages/kmod-6.1-nvidia-r580/grid-license-check.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=GRID License Check
+RefuseManualStart=true
+RefuseManualStop=true
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+ExecStart=-/usr/bin/truncate /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/nvidia-smi -q
+ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
+StandardOutput=append:/tmp/.nvidia-gridd-license
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r580/grid-license-check.timer
+++ b/packages/kmod-6.1-nvidia-r580/grid-license-check.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=GRID License Check Timer
+RefuseManualStart=true
+
+[Timer]
+Unit=grid-license-check.service
+OnBootSec=5s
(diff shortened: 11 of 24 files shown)


packages/kmod-6.1-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,26 @@
...
+RefuseManualStop=true
+DefaultDependencies=no
+Before=kubelet.service
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+# Otherwise, attempt to load the module.
```
arnaldo2792: I'm confused, will this line actually load the module or are you just generating output that will be `grep`ed later?

packages/kmod-6.12-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,26 @@
...
+Before=kubelet.service
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+# Otherwise, attempt to load the module.
+ExecStart=/usr/bin/nvidia-smi -q
+# Ensure that the stderr file exists. Otherwise, grep fails on an empty file.
```
arnaldo2792: It's the STDOUT file what you are creating, right?

packages/kmod-6.1-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,26 @@
...
+# Succeed unless there was a fatal error.
+ExecStart=/usr/bin/grep -Fqvzw Unlicensed /tmp/.nvidia-gridd-license
+RemainAfterExit=true
+StandardOutput=append:/tmp/.nvidia-gridd-license
+Restart=on-failure
+RestartSec=1
+StartLimitBurst=120
+
+[Install]
+RequiredBy=kubelet.service
```
bcressey: Below it's required by `nvidia-k8s-device-plugin.service` which seems more correct.

packages/kmod-6.1-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,26 @@
...
+After=nvidia-gridd.service
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+# Otherwise, attempt to load the module.
+ExecStart=/usr/bin/nvidia-smi -q
+# Ensure that the stderr file exists. Otherwise, grep fails on an empty file.
+ExecStart=-/usr/bin/touch /tmp/.nvidia-gridd-license
```
bcressey: `truncate` may be better than `touch` to reset the file between iterations

(would have to move this before the `nvidia-smi` call)

packages/kmod-6.12-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,26 @@
...
+# Succeed unless there was a fatal error.
+ExecStart=/usr/bin/grep -Fqvzw Unlicensed /tmp/.nvidia-gridd-license
+RemainAfterExit=true
+StandardOutput=append:/tmp/.nvidia-gridd-license
+Restart=on-failure
+RestartSec=1
+StartLimitBurst=120
+
+[Install]
+RequiredBy=nvidia-k8s-device-plugin.service
```
bcressey: Does this requirement cause the k8s device plugin to fail if the license check fails? I'm kind of worried about cluttering up the logs with a lot of failures.

This could possibly be modeled as:
1. a timer unit that runs and creates a marker file when the license check passes
2. a path unit that activates `nvidia-k8s-device-plugin.service`
3. a fallback unit that runs when we don't match the grid driver that also creates the marker
4. a condition in the k8s device plugin that requires the marker to exist

packages/kmod-6.1-nvidia-r570/grid-license-check.service
yeazelm: This comment might make more sense above the ExecCondition
```suggestion
# Confirm GRID is required, then attempt to check the license
```

packages/kmod-6.1-nvidia-r570/grid-license-check.service
yeazelm: I like that, I'll give it a shot to see if it still gives me the behavior I want with things rearranged to use `truncate`

packages/kmod-6.1-nvidia-r570/grid-license-check.service
yeazelm: Sorry, this was originally `nvidia-k8s-device-plugin.service` and I missed the refactor. In my mind we have several options for what experience we want:
* Don't prevent anything but log it. Sort of the behavior we have today just without logs (not ideal)
* Prevent kubelet from starting if the license isn't valid. This means the node never becomes ready and cannot accept any orchestrated work. (This is what I was intending to enforce in the PR)
* Prevent the device plugin from running, the node will become ready but not advertise GPU resources. It could take some work but not GPU work. (This might be confusing to users so I leaned away from this but was my original approach)
* Attempt to block boot entirely so the node doesn't even reach multi-user. (Seems harsh but would be fine too).

packages/kmod-6.12-nvidia-r570/grid-license-check.service
yeazelm: Yes, this is a forgotten update, I moved to STDOUT but forgot to update the comment.

packages/kmod-6.12-nvidia-r570/grid-license-check.service
yeazelm: This might be much cleaner in the logs, otherwise the unit is very angry and noisy in the journal when its failing. I'll play with that as a potential alternative to this. FWIW though I haven't seen this fail yet before the next unit runs when we are going to get a license, so it might be a situation where the only time its noisy, is when the node is already in a bad state. Nonetheless, I think making it cleaner is worth it.

yeazelm:
^ I updated the approach to use timers and a marker file. I'm working on all the new logs from testing to confirm the behavior works as described.

packages/kmod-6.1-nvidia-r570/grid-license-check.service
yeazelm: I decided to go with nvidia-k8s-device-plugin.service so the instance comes up and is able to be investigated but won't advertise GPUs.

packages/kmod-6.12-nvidia-r570/grid-license-check.service
yeazelm: This is the new approach.

packages/kmod-6.1-nvidia-r570/grid-license-check.timer
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=GRID License Check Timer
+RefuseManualStart=true
+DefaultDependencies=no
```
bcressey: Why `DefaultDependencies=no`? 
```suggestion
RefuseManualStart=true
```

We need to be careful that this timer doesn't queue a start job for something ahead of schedule. In this case, it could cause all the dependencies of `grid-license-check.service` to start before we're ready, at least on FIPS variants where we start in `fipscheck.target` rather than `preconfigured.target`.

packages/kmod-6.1-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,20 @@
...
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+ExecStart=-/usr/bin/truncate /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/nvidia-smi -q
+ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer
+StandardOutput=append:/tmp/.nvidia-gridd-license
+
+[Install]
+WantedBy=nvidia-k8s-device-plugin.service
```
bcressey, line 19: Do we need this dependency?

packages/kmod-6.1-nvidia-r570/tesla-license-fallback.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+DefaultDependencies=no
+
+[Service]
+Type=oneshot
+ExecCondition=!/usr/bin/ghostdog match-nvidia-driver tesla
```
bcressey: I'm not clear on what this is meant to be doing.

`ExecCondition` doesn't support negation, and if it did, this would mean "execute if we're not using the tesla driver" which seems like the opposite of what we want.

packages/kmod-6.1-nvidia-r570/grid-license-check.service
```diff
@@ -0,0 +1,20 @@
...
+Requires=nvidia-gridd.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver grid
+ExecStart=-/usr/bin/truncate /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/nvidia-smi -q
+ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-license
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer
```
bcressey: I'd always recommend `--no-block` unless you're very sure this could never possibly deadlock:
```suggestion
ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
```

bcressey (commented):
For this:
> A change will need to be added to the nvidia-k8s-device-plugin.service to depend on this file

Can you add this as a drop-in so there's no cross-kit dependency?

yeazelm:
^ Updated for the comments and added in the drop-in for nvidia-k8s-device-plugin.

packages/kmod-6.1-nvidia-r570/grid-license-check.service
yeazelm: I moved to the drop in

packages/kmod-6.12-nvidia-r580/tesla-license-fallback.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/ghostdog match-nvidia-driver tesla
+ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
+ExecStart=/usr/bin/systemctl stop grid-license-check.timer
```
bcressey:
```suggestion
ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
```

packages/kmod-6.1-nvidia-r580/tesla-license-fallback.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Tesla GRID License Check Fallback
+RefuseManualStart=true
+RefuseManualStop=tru
```
bcressey:
```suggestion
RefuseManualStop=true
```

bcressey (approved):
Two minor fixes, LGTM otherwise.

yeazelm:
^ Fixed typos from comments/

packages/kmod-6.1-nvidia-r570/grid-license-file-check.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPre=/usr/bin/test -f /etc/drivers/.grid-licensed
```
arnaldo2792, line 2: Why not `ConditionPathExists`?

https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionPathExists=

arnaldo2792 (approved):

packages/kmod-6.1-nvidia-r570/grid-license-file-check.conf
yeazelm: `ConditionPathExists` will fail the unit, I want systemd to retry. If the unit happens to get scheduled before we have placed this file, it would just mark it as failed. The `ExecStartPre` matches the rest of the services dependencies. See https://github.com/bottlerocket-os/bottlerocket-core-kit/blob/develop/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service for the other pre conditions.

PR #322, opened 2025-11-17
Test EFA configurations in parallel

arnaldo2792:
**Description of changes:**

In e0ab5af, a patch was introduced to execute EFA's configuration validations in serial. The problem the patch fixed was self-inflicted by the patch 1001-Makefile-add-prepare-target-for-external-modules.patch, which is required in Bottlerocket to support building out of tree kernel modules.

Update said patch to allow the parallel EFA configuration validations, and drop the patch to make the validations serial.

**Testing done:**

Builds successfully, and internal testing confirmed the driver is functional

**Terms of contribution:**


yeazelm (approved):

arnaldo2792:
(Forced push actually drops the `efa.cmake` patch)

arnaldo2792:
(Forced push to rebase onto develop)

bcressey (approved):

PR #323, opened 2025-11-18
Extend supported MIG profiles

arnaldo2792:
**Description of changes:**

This change adds support for NVIDIA's B300 cards. The card supports 6 types of profiles, however for profiles `4g.135gb` and `7g.269gb`  the card uses the same alias (`1`) as the number of partitions created for the profiles. I used `7g.269gb` for the `1` alias since in Bottlerocket we use `1` for the profile that uses the entire GPU.

**Testing done:**

I tested all the profiles and confirmed the GPU was partitioned as expected.

**Terms of contribution:**

### DIFF
--- a/sources/nvidia-migmanager/src/mig_profile.rs
+++ b/sources/nvidia-migmanager/src/mig_profile.rs
@@ -1,5 +1,9 @@
 use serde::Deserialize;
 
+pub(crate) trait MigGpuProfile: for<'de> Deserialize<'de> {
+    fn get_mig_profile(&self) -> &str;
+}
+
 #[derive(Deserialize)]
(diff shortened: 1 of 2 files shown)


sources/nvidia-migmanager/src/mig_profile.rs
```diff
@@ -1,5 +1,9 @@
 use serde::Deserialize;
 
+pub(crate) trait MigGpuProfile: for<'de> Deserialize<'de> {
```
arnaldo2792, line 3: I moved this here because we kept punting it down and that polluted the actual changes. Now going forward we can just add changes at the bottom of this file.

yeazelm (approved):

piyush-jena (approved):

Issue #410, opened 2026-04-21
Cross-compilation of the 6.18 kernel fails


vigh-m:
**Platform I'm building on:**
`aarch64`

**What I expected to happen:**
Successful kernel-kit

**What actually happened:**
Kernel-kit build failed with the following for the 6.18 kernel
```
   > [rpmbuild 6/7] RUN --mount=source=.cargo/twoliter_cargo_config.toml,target=/home/builder/rpmbuild/.cargo/config.toml     --mount=source=.cargo/vendor,target=/home/builder/rpmbuild/.cargo/vendor     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources     --mount=target=/host     /host/build/tools/unplug       rpmbuild -bb --clean         --undefine _auto_set_build_flags         --define "_target_cpu x86_64"         --define "dist .1776123295.6965cb0d.br1"         rpmbuild/SPECS/kernel-6.18.spec:
  42.76 406a407,408
  42.76 > CONFIG_CC_HAS_MARCH_NATIVE=y
  42.76 > # CONFIG_X86_NATIVE_CPU is not set
  42.76 + echo 'error: source and build kernel configurations do not match'
  42.76 + exit 1
  42.76 error: source and build kernel configurations do not match
  42.77
  42.77 RPM build errors:
  42.77 error: Bad exit status from /var/tmp/rpm-tmp.dZ46zD (%prep)
  42.77     Bad exit status from /var/tmp/rpm-tmp.dZ46zD (%prep)
```

**How to reproduce the problem:**

- Checkout the latest kernel-kit v5.3.3 releaese
- Cross compile by building the x86_64 kit on an aarch64 instance

Issue #409, opened 2026-04-21
Unable to run `make full-config` on an ARM instance


vigh-m:
**Platform I'm building on:**
ARM

**What I expected to happen:**
`make full-config` will generate the kernel configs

**What actually happened:**
```bash
$ make full-config
SDK="" ./tools/docker-run.sh "/bottlerocket-kernel-kit/tools/latest-kernel-full-config.sh"
Retriving SDK from Twoliter.lock
Unable to find image 'ghcr.io/mdm-code/tq:v2.4.0' locally
v2.4.0: Pulling from mdm-code/tq
21cec107bac3: Pull complete
Digest: sha256:68802505ce28adb912005c17e09723ad43c4e49067b6943df1171e71012805a4
Status: Downloaded newer image for ghcr.io/mdm-code/tq:v2.4.0
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /bin/tq: exec format error
make: *** [Makefile:26: full-config] Error 255
```

**How to reproduce the problem:**

PR #411, opened 2026-04-22
Fix kernel 6.18 build

vigh-m:
**Issue number:**
Closes 
- https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/409
- https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/410

**Description of changes:**
- Move from using `tq` to `yq` to parse the `Twoliter.lock` toml file.  `yq` has ARM support. `make full-config` now works on both x86 and aarch64 hosts
- Build architecture and host specific configs for kernel-6.18. This unblocks the cross-compilation of kernel-6.18 on aarch64 hosts. 
The file format chosen here is `config-full-bottlerocket-{target-arch}-on-{host-arch}` so the aarch64 config built on an x86 instance would be `config-full-bottlerocket-aarch64-on-x86_64`

The difference to look out for with this change is `CONFIG_CC_HAS_MARCH_NATIVE` and `CONFIG_X86_NATIVE_CPU` are no longer present in `config-full-bottlerocket-x86_64-on-aarch64`

The change is limited to the 6.18 version of the kernel as that is the only kernel that requires this split right now. The code change should support adding other kernels as needed in the future

**Testing done:**
-  Built the kernel-kit with all 4 combinations of hosts and arches.

**Terms of contribution:**


tools/latest-kernel-full-config.sh
```diff
@@ -39,6 +39,17 @@ bail() {
     exit 1
 }
 
+# Get the config-full filename for a given kernel version and architecture
+get_kernel_config_file_name() {
+    local majorminor="$1"
+    local arch="$2"
+    if [ "${majorminor}" == "6.18" ]; then
```
jpculp:
```suggestion
    if [[ "${majorminor}" == "6.18" ]]; then
```
Better to have double brackets with double equals as `[...]` is a POSIX test while `==` is a Bashism.

jpculp (dismissed):

tools/latest-kernel-full-config.sh
```diff
@@ -39,6 +39,17 @@ bail() {
...
+        echo "config-full-bottlerocket-${arch}-on-$(uname -m)"
+    else
+        echo "config-full-bottlerocket-${arch}"
+    fi
+}
```
ytsssun, line 51: Nit: could we add some comments on why we did this only for 6.18 for future context?

vigh-m: Yup. I'll add context to the PR description.

vigh-m:
⬆️ Removed the binutils change to let the CI build

vigh-m:
⬆️ Addressed comments. Also ran `shfmt` to fixup the formatting

mgsharm:
`packages/kernel-6.18/README.md` needs updating, it still describes the old 2‑file layout.

vigh-m:
⬆️ Updated `kernel-6.18/README`.

ytsssun (approved):

mgsharm (approved):

KCSesh:
LGTM

KCSesh (approved):

PR #419, opened 2026-05-04
docker-run: replace yq with grep/cut

jpculp:
**Description of changes:**

To reduce dependencies on external tooling and repositories, simplify the SDK source fetch using core utilities such as `grep` and `cut`.

**Testing done:**

`make full-config`

**Terms of contribution:**

### DIFF
--- a/tools/docker-run.sh
+++ b/tools/docker-run.sh
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -e -o pipefail
 
-YQ_IMAGE=mikefarah/yq:4.53.2
-
 bail() {
     if [[ $# -gt 0 ]]; then
         >&2 echo "Error: $*"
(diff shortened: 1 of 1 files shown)


tools/docker-run.sh
```diff
@@ -11,12 +9,7 @@ bail() {
 }
 
 find_sdk() {
```
jpculp, line 11: I figured we could keep this one-liner wrapped in a `find_sdk` function just in case our needs become complex and we need to bring back in more advanced logic.

ginglis13 (approved):

tools/docker-run.sh
```diff
@@ -11,12 +9,7 @@ bail() {
...
-  docker run --rm \
-      -v "$(pwd):/bottlerocket-kernel-kit:ro" \
-      --user "$(id -u):$(id -g)" \
-      --network none \
-      "${YQ_IMAGE}" \
-      -p toml -oy '.sdk.source' /bottlerocket-kernel-kit/Twoliter.lock
+  grep -A5 '^\[sdk\]' Twoliter.lock | grep '^source' | cut -d'"' -f2
```
vigh-m, line 12: 1. This is a nit but we can achieve the same with `A4`
2. Why did the path change from using `${pwd}` to just `Twoliter.lock`. Should we keep the `${pwd}` bit?

jpculp: 1. We can, but given that technically the whole block is A5. I suspect it's safer to take in the whole block just in case the order were to ever shuffle.
2. Since we're not going into a container to accomplish this we can just refer to the path relatively. We could make this more robust with something like `repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"`, but if you were to run `make full-config` in a different directory I think it would still fail because it is also using relative paths.

vigh-m (approved):

PR #425, opened 2026-05-05
Normalize NVIDIA library paths to /usr/lib/

maherthomsi:
Merge with https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/919
**Description of changes:**

Normalize NVIDIA library installation paths from `/usr/lib/nvidia/tesla/` to the standard `/usr/lib/` path. Create backwards-compatibility symlinks in `/usr/lib/nvidia/tesla/` pointing back to `/usr/lib/` for workloads thatreference the legacy path.

Changes across kmod-6.1-nvidia-r580, kmod-6.12-nvidia-r580, and kmod-6.18-nvidia-r580:

  - Move library installation from `nvidia/tesla/` subdirectory to `/usr/lib/` directly
  - Add `%{_cross_libdir}/nvidia/tesla/*` to `%files` for the backwards-compat symlinks
  - Remove `nvidia-ld.so.conf.in` (Source302) and all references — no longer needed since libraries are in the standard linker search path
  - Remove `nvidia/tesla` tmpfiles directory entry

**Testing done:**

  - Built kernel-kit for both x86_64 and aarch64
  - Published to ECR and built aws-k8s-1.35-nvidia variant AMIs
  - Launched g4dn.xlarge (Tesla T4) node on EKS cluster
  - Verified nvidia-smi works (driver 580.126.09)
  - Verified on host: libraries at `/usr/lib/`, symlinks at `/usr/lib/nvidia/tesla/` → `../../<lib>`
  - Verified in container: libraries mounted at `/usr/lib/`, symlinks at `/usr/lib/nvidia/tesla/` → `/usr/lib/<lib>`
  - Ran nvidia smoke tests - all passed
  
**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
+++ b/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
@@ -58,7 +58,6 @@ Source214: grid-license-file-check.conf
 # NVIDIA tesla conf files from 300 to 399
 Source300: nvidia-tesla-tmpfiles.conf
 Source301: nvidia-tesla-build-config.toml.in
-Source302: nvidia-ld.so.conf.in
 
 # Driverdog config templates from 400 to 499
 Source400: nvidia-open-gpu-config.toml.in
@@ -247,9 +246,9 @@ popd
 
 %install
 install -d %{buildroot}%{_cross_libdir}
+install -d %{buildroot}%{_cross_libdir}/nvidia/tesla
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -d %{buildroot}%{_cross_unitdir}
-install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/{drivers,ld.so.conf.d}
 install -d %{buildroot}%{_cross_sysusersdir}
 install -d %{buildroot}%{_cross_bindir}
 
@@ -266,6 +265,7 @@ install -p -m 0644 %{S:202} %{buildroot}%{_cross_libdir}/modules-load.d/nvidia-d
 # NVIDIA fabric manager service unit and config
 install -p -m 0644 %{S:203} %{buildroot}%{_cross_unitdir}
 install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/drivers
 install -p -m 0644 %{S:204} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.cfg
 install -p -m 0644 %{S:207} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia/fabricmanager.env
 
@@ -274,7 +274,6 @@ pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}
 # Proprietary driver
 install -d %{buildroot}%{_cross_bindir}
 install -d %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin
-install -d %{buildroot}%{_cross_libdir}/nvidia/tesla
 install -d %{buildroot}%{_cross_datadir}/nvidia/tesla/module-objects.d
 install -d %{buildroot}%{_cross_factorydir}/nvidia/tesla
 install -d %{buildroot}%{_cross_factorydir}/nvidia/open-gpu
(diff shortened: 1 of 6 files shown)


mgsharm:
Can we have separate commits for each kmod. Also the commit is not signed and doesn't have Signed-off-by.

maherthomsi:
Split up commits for each kmod. Signed commits and added `Signed off by` to commits

maherthomsi:
Force push to fix merge conflicts

jpculp (approved):

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
```diff
@@ -481,6 +480,8 @@ popd
 %{_cross_attribution_file}
 %dir %{_cross_libexecdir}/nvidia
 %dir %{_cross_libdir}/nvidia
+%dir %{_cross_libdir}/nvidia/tesla
+%{_cross_libdir}/nvidia/tesla/*
```
mgsharm, line 481: Can we list the files explicitly instead of using *?

The `%{_cross_libdir}/nvidia/tesla/*` glob removes a safety check we rely on today. Right now every .so is listed by name, and the install step in spec file has this comment:

```
  # We install all the libraries, and filter them out in the 'files' section,
  # so we can catch when new libraries are added
```

If NVIDIA adds a new library in a future driver, the build fails on unpackaged files and we get to decide whether to ship it. With tesla/*, anything new silently lands in the image and %exclude is the only thing in the way.

This applies to other spec file in the review.

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
```diff
@@ -449,7 +443,12 @@ for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%%P\n'); do
   soname="$(%{_cross_target}-readelf -d "${lib}" | awk '/SONAME/{print $5}' | tr -d '[]')"
   [ -n "${soname}" ] || continue
   [ "${lib}" == "${soname}" ] && continue
-  ln -s "${lib}" %{buildroot}/%{_cross_libdir}/nvidia/tesla/"${soname}"
+  ln -s "${lib}" %{buildroot}/%{_cross_libdir}/"${soname}"
+done
+
+# Create per-library symlinks in nvidia/tesla/ for backwards compatibility
+for lib in $(find %{buildroot}/%{_cross_libdir} -maxdepth 1 -type f -name '*.so*' -printf '%%P\n'); do
```
arnaldo2792: You could move this to the for above, before the `soname` check.

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
arnaldo2792: This glob should be fine to keep. The expanded list is now in `%{_cross_libdir}` so we won't miss libraries

packages/kmod-6.12-nvidia-r580/kmod-6.12-nvidia-r580.spec
```diff
@@ -468,7 +462,12 @@ for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%%P\n'); do
   soname="$(%{_cross_target}-readelf -d "${lib}" | awk '/SONAME/{print $5}' | tr -d '[]')"
   [ -n "${soname}" ] || continue
   [ "${lib}" == "${soname}" ] && continue
-  ln -s "${lib}" %{buildroot}/%{_cross_libdir}/nvidia/tesla/"${soname}"
+  ln -s "${lib}" %{buildroot}/%{_cross_libdir}/"${soname}"
+done
+
+# Create per-library symlinks in nvidia/tesla/ for backwards compatibility
+for lib in $(find %{buildroot}/%{_cross_libdir} -maxdepth 1 -type f -name '*.so*' -printf '%%P\n'); do
```
arnaldo2792: Same as elsewhere, you could include this line in the for above.

maherthomsi:
Force push:  Addresses above comments to move symlink creation into the existing for-loop across all three spec files (6.1, 6.12, 6.18)

mgsharm (approved):

vigh-m (approved):

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
```diff
@@ -446,10 +440,12 @@ EXCLUDED_LIBS="libEGL.so.%{tesla_ver}"
 
 for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%%P\n'); do
   [[ "${EXCLUDED_LIBS}" =~ "${lib}" ]] && continue
+  # Create backwards-compatibility symlink in nvidia/tesla/
+  ln -s "../../${lib}" %{buildroot}/%{_cross_libdir}/nvidia/tesla/"${lib}"
```
arnaldo2792: Lets follow the approach elsewhere in the repo, and use `ls -rs` with absolute paths:

```spec
ls -rs "%{buildroot}/%{_cross_libdir}/${lib}" "%{buildroot}/%{_cross_libdir}/nvidia/tesla/${lib}"
```

arnaldo2792 (approved):

PR #442, opened 2026-05-20
feat: enable CONFIG_SQUASHFS=m for kernel-6.12 and kernel-6.18

### LINKED ISSUE #4845: add support for squashfs kernel module
**What I'd like:**

We're evaluating migrating some Kubernetes-based workloads over to AWS and running them on EKS, and would prefer using Bottlerocket Linux as our host OS.

These existing workloads do a lot of work with squashfs loop-mounts. The squashfs kernel-module does not seem to be available on bottlerocket linux - when I attempt to create a mount from a privileged Kubernetes pod I get the following error:

```
$ mount -t squashfs -o loop /something.sqfs /mount/location
mount: /mount/location: unknown filesystem type 'squashfs'.
dmesg(1) may have more information after failed mount system call.
```

(dmesg did not have more information)

```
$ uname -a
Linux pod-name 6.12.79 #1 SMP PREEMPT_DYNAMIC Fri May  1 00:25:24 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
```

Node-image-version: `v1.34.4-eks-f69f56f`

- FUSE-mounted squashfs
- AL2023 host OS

We will likely not pursue our own custom builds of Bottlerocket, but I admit I haven't looked into it yet.

If the kernel-module had to be disabled-by-default with a flag to enable it in user data that wouldn't be a problem.

ginglis13:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket/issues/4845

**Description of changes:**

Enable the squashfs kernel module so that userspace applications
can mount squashfs filesystems. This restores the squashfs support
that was present in kernel-6.1 but dropped in newer kernels.

Configs are copied as the kernel-6.1 configs for squashfs: https://github.com/bottlerocket-os/bottlerocket-kernel-kit/blob/e9e21d225845e9e3866d515893f37e8194832fba/packages/kernel-6.1/config-bottlerocket#L14-L29 All configs added with a value of `y` are boolean, hence why they are not `m`.

 Only `config-bottlerocket` files generated by hand. All other configs generated with `make full-config` ran on both x86 and aarch64 hosts.

**Testing done:**

Pod for testing:

<details>

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-squashfs
  namespace: default
spec:
  containers:
  - name: test
    image: public.ecr.aws/amazonlinux/amazonlinux:2023
    command:
    - /bin/bash
    - -c
    - |
      set -x

      echo "=== Installing tools ==="
      yum install -y util-linux kmod squashfs-tools procps-ng > /dev/null 2>&1

      echo ""
      echo "=== Kernel version ==="
      uname -r

      echo ""
      echo "=== Check if squashfs module is available ==="
      modprobe squashfs && echo "modprobe squashfs: OK" || echo "modprobe squashfs: FAILED"

      echo ""
      echo "=== Verify module is loaded ==="
      lsmod | grep squashfs

      echo ""
      echo "=== Check /proc/filesystems for squashfs ==="
      cat /proc/filesystems | grep squashfs

      echo ""
      echo "=== Test from issue #4845: mount a dummy squashfs file ==="
      echo "Expected: 'wrong fs type, bad option, bad superblock' (means kernel recognizes squashfs)"
      echo "Failure:  'unknown filesystem type squashfs' (means module not available)"
      for i in 0 1 2 3 4 5 6 7; do mknod -m 660 /dev/loop$i b 7 $i 2>/dev/null || true; done
      truncate -s 4K /tmp/f.sqfs && mkdir -p /tmp/m
      mount -t squashfs -o loop /tmp/f.sqfs /tmp/m 2>&1; echo "exit code: $?"

      echo ""
      echo "=== Test with a real squashfs image ==="
      mkdir -p /tmp/squashfs-root
      echo "hello from squashfs" > /tmp/squashfs-root/hello.txt
      mksquashfs /tmp/squashfs-root /tmp/real.sqfs -noappend -quiet
      mkdir -p /tmp/mnt
      mount -t squashfs -o loop /tmp/real.sqfs /tmp/mnt
      echo "Mount exit code: $?"
      echo "Contents:"
      cat /tmp/mnt/hello.txt
      umount /tmp/mnt

      echo ""
      echo "=== Test compression algorithms ==="
      for algo in gzip lz4 lzo xz zstd; do
        mksquashfs /tmp/squashfs-root /tmp/test-${algo}.sqfs -noappend -quiet -comp ${algo} 2>/dev/null
        if mount -t squashfs -o loop /tmp/test-${algo}.sqfs /tmp/mnt 2>/dev/null; then
          echo "${algo}: OK ($(cat /tmp/mnt/hello.txt))"
          umount /tmp/mnt
        else
          echo "${algo}: FAILED"
        fi
      done

      echo ""
      echo "=== dmesg squashfs entries ==="
      dmesg | grep -i squashfs | tail -10

      echo ""
      echo "=== DONE ==="
      sleep infinity
    securityContext:
      privileged: true
    volumeMounts:
    - name: modules
      mountPath: /lib/modules
      readOnly: true
  volumes:
  - name: modules
    hostPath:
      path: /lib/modules
  restartPolicy: Never
  nodeSelector:
    kubernetes.io/arch: amd64
```
</details>

`aws-k8s-1.36-x86_64` (kernel-6.18) and `aws-k8s-1.35-x86_64` (kernel-6.12) used for test with the following results:

 | | Kernel 6.12 (k8s-1.35) | Kernel 6.18 (k8s-1.36) |
  |--|--|--|
  | `modprobe squashfs` | OK | OK |
  | Issue #4845 test | correct "bad superblock" | correct "bad superblock" |
  | Real mount | OK | OK |
  | gzip | OK | OK |
  | lz4 | OK | OK |
  | lzo | OK | OK |
  | xz | OK | OK |
  | zstd | OK | OK |

**Terms of contribution:**


ginglis13:
^ force push moves the module to be in alphabetical order and brings the squashfs config to the top of `config-bottlerocket`

packages/kernel-6.12/config-bottlerocket
```diff
@@ -4,6 +4,25 @@
 # The root filesystem is erofs
 CONFIG_EROFS_FS=y
 
+# Some userspace applications require squashfs support
```
arnaldo2792, line 7: Your commit doesn't follow the convention for the commits in the kernel kit. You should also add your `Sign-Off` in the commit.

Also, please split the commits, one per kernel.

packages/kernel-6.12/config-bottlerocket
```diff
@@ -4,6 +4,25 @@
...
+# CONFIG_SQUASHFS_CHOICE_DECOMP_BY_MOUNT is not set
+CONFIG_SQUASHFS_COMPILE_DECOMP_SINGLE=y
+# CONFIG_SQUASHFS_COMPILE_DECOMP_MULTI is not set
+# CONFIG_SQUASHFS_COMPILE_DECOMP_MULTI_PERCPU is not set
+CONFIG_SQUASHFS_XATTR=y
+CONFIG_SQUASHFS_ZLIB=y
+CONFIG_SQUASHFS_LZ4=y
+CONFIG_SQUASHFS_LZO=y
+CONFIG_SQUASHFS_XZ=y
+CONFIG_SQUASHFS_ZSTD=y
```
arnaldo2792, line 21: Can you please comment in the PR description that all these can't be `m` because they are boolean configurations ?

ginglis13: updated PR description

arnaldo2792 (approved):

mgsharm (approved):

PR #451, opened 2026-05-27
Bake in virtio and NSM kernel modules in kernel image

arnaldo2792:
**Description of changes:**

Provide the virtio and NSM kernel modules in the kernel image, instead of loadable modules. They are needed for virtualized environments like MicroVMs or Nitro Enclaves, since Bottlerocket doesn't provide a bootable initrd where the modules can be loaded from.

**Terms of contribution:**


ginglis13 (approved):

piyush-jena (approved):

arnaldo2792:
We are going to go in a different direction for now.

PR #475, opened 2026-06-26
Fix nvidia fallback services failing to start

### LINKED ISSUE #464: [bug] open-gpu-license-fallback.service and tesla-license-fallback.service never run on non-Kubernetes hosts due to k8s-only WantedBy
**Package I'm using:**
The NVIDIA `kmod-*-nvidia-r580` packages in bottlerocket-kernel-kit. The same files
exist with the same `[Install]` configuration in `kmod-6.1-nvidia-r580`,
`kmod-6.12-nvidia-r580`, and `kmod-6.18-nvidia-r580`, so this applies to all three
(and likely the r595 variants too).
**What I expected to happen:**
On open-gpu hosts (e.g. NVIDIA L4 with the open-gpu kernel modules), the kit's
`open-gpu-license-fallback.service` should run and `touch /etc/drivers/.grid-licensed`,
after which `grid-license-check.service` no-ops on every subsequent timer fire — a
quiet system. The GRID licensing stack is for hosts running the proprietary `grid`
driver, not open-gpu. Equivalently for tesla driver hosts via
`tesla-license-fallback.service`.
**What actually happened:**
On a non-Kubernetes Bottlerocket consumer the fallback **never runs at all**:
```
$ systemctl status open-gpu-license-fallback.service
○ open-gpu-license-fallback.service - Open GPU GRID License Check Fallback
     Loaded: loaded (...; enabled; preset: enabled)
     Active: inactive (dead)
$ ls -l /etc/drivers/.grid-licensed
ls: cannot access ...: No such file or directory
```
Meanwhile `grid-license-check.timer` (`OnUnitActiveSec=2s`) fires every 2 seconds,
restarting `grid-license-check.service` and pulling `nvidia-gridd.service` along via
`Requires=`. Each cycle's services do no real work on open-gpu hosts (they're gated
to the grid driver), but the systemd churn is real — ~100 activations per service per
5 minutes — and contends for the systemd main thread. Downstream we observed it as
elevated tail latency on unrelated systemd/D-Bus operations on GPU hosts.
Root cause — the two units have asymmetric `[Install]` targets:
```
# grid-license-check.timer
[Install]
WantedBy=timers.target                        # universal — reached on every boot
# open-gpu-license-fallback.service
[Install]
WantedBy=nvidia-k8s-device-plugin.service     # only pulled in on k8s nodes
```
So on any non-Kubernetes consumer the timer is enabled and fires forever, while its
silencer is never pulled into the boot transaction. `tesla-license-fallback.service`
has the identical k8s-only `WantedBy` and fails the same way for tesla driver hosts.
**How to reproduce the problem:**
1. Boot a non-Kubernetes Bottlerocket image with one of the `kmod-*-nvidia-r580`
   packages on open-gpu-preferred hardware (e.g. an L4-class instance).
2. `systemctl status open-gpu-license-fallback.service` → `inactive (dead)`.
3. `ls /etc/drivers/.grid-licensed` → does not exist.
4. `journalctl -u grid-license-check.service --since "5 min ago" | grep -c "Starting"` → ~100.
---
**Suggested fix:**
Anchor the license fallbacks to a universally-reached target so they run on every
boot, matching the timer's reachability. In each of `kmod-6.1-nvidia-r580`,
`kmod-6.12-nvidia-r580`, and `kmod-6.18-nvidia-r580` (and their r595 counterparts):
```diff
 # open-gpu-license-fallback.service
 [Install]
-WantedBy=nvidia-k8s-device-plugin.service
+WantedBy=multi-user.target
```
…and the same for `tesla-license-fallback.service`. The existing
`ExecCondition=ghostdog match-nvidia-driver <flavor>` keeps each unit a no-op on
hosts not running that driver, so this is safe across all GPU flavors. This trades
steady-state spam for a small number of early-boot activations until the fallback
runs; if maintainers prefer, an `After=`/`Before=` ordering or anchoring earlier
(e.g. `sysinit.target`) would tighten that further. An alternative would be to gate
the timer behind the same k8s target so neither runs off-Kubernetes, but that loses
the licensing path on k8s hosts where the fallback is still wanted; the `[Install]`
change is the simpler one.
**Workaround:**
Until this is fixed upstream, downstream consumers can ship a separate oneshot that
masks `nvidia-gridd.service`, `grid-license-check.service`, and
`grid-license-check.timer` at boot, gated by
`ExecCondition=/usr/bin/ghostdog match-nvidia-driver open-gpu`. Happy to remove that
local workaround once the `[Install]`-target fix lands here.

Happy to send a PR if maintainers agree on the approach.

vigh-m:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/464

The `open-gpu-license-fallback.service` and the `tesla-license-fallback.service` were set to start on `nvidia-k8s-device-plugin.service` and thus would not trigger on non-k8s variants. This caused the `grid-license-check.service` to never turn off, filling the journal log with spurious messages.

**Description of changes:**
- Add the `%build` sections to kmod-nvidia specfiles
- Fix type in DevID for `g4dn` instances causing the open driver to be loaded instead of the tesla driver. With this change, we are in-line with what the EKS optimized AMI does
- Change the dependency of the `open-gpu-license-fallback.service` and the `tesla-license-fallback.service` to depend on the `nvidia-gridd.service` which itself depends on `preconfigured.target` ensuring all are started on all variants.

**Testing done:**
- Build and test the a combination of ecs and k8s variants on the 6.1, 6.12, and 6.18 kernels with the 580 driver

- Testing for `g4dn.*` instances:
Before the change:
```bash
bash-5.2# nvidia-smi
Wed Jul  8 00:40:56 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.159.03             Driver Version: 580.159.03     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  Tesla T4                       On  |   00000000:00:1E.0 Off |                    0 |
| N/A   36C    P8              9W /   70W |       0MiB /  15360MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
bash-5.2# modinfo nvidia
filename:       /lib/modules/6.18.30/kernel/drivers/extra/video/nvidia/open-gpu/nvidia.ko
import_ns:      DMA_BUF
alias:          char-major-195-*
description:    NVIDIA core GPU kernel module
version:        580.159.03
supported:      external
license:        Dual MIT/GPL    <------- Open Source Driver

bash-5.2# /usr/bin/ghostdog match-nvidia-driver tesla
Error: tesla is not preferred driver: open-gpu

bash-5.2# /usr/bin/ghostdog match-nvidia-driver grid
Error: grid is not preferred driver: open-gpu

bash-5.2# /usr/bin/ghostdog match-nvidia-driver open-gpu
bash-5.2#
```

After the change
```bash
bash-5.2# nvidia-smi
Wed Jul  8 00:36:59 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.159.03             Driver Version: 580.159.03     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  Tesla T4                       On  |   00000000:00:1E.0 Off |                    0 |
| N/A   35C    P8             10W /   70W |       0MiB /  15360MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
bash-5.2# modinfo nvidia
filename:       /lib/modules/6.18.35/kernel/drivers/extra/video/nvidia/tesla/nvidia.ko
alias:          char-major-195-*
description:    NVIDIA core GPU kernel module
version:        580.159.03
supported:      external
license:        NVIDIA    <------- Proprietary Driver


bash-5.2# /usr/bin/ghostdog match-nvidia-driver tesla
bash-5.2# /usr/bin/ghostdog match-nvidia-driver grid
Error: grid is not preferred driver: tesla

bash-5.2# /usr/bin/ghostdog match-nvidia-driver open-gpu
Error: open-gpu is not preferred driver: tesla
```

**Terms of contribution:**

### DIFF
--- a/packages/kmod-6.1-nvidia-r580/grid-license-check.service
+++ b/packages/kmod-6.1-nvidia-r580/grid-license-check.service
@@ -14,6 +14,3 @@ ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-licen
 ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
 ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 StandardOutput=append:/tmp/.nvidia-gridd-license
-
-[Install]
-WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
+++ b/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
@@ -188,6 +188,8 @@ tar -xf %{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
 # This recipe was based in the NVIDIA yum/dnf specs:
 # https://github.com/NVIDIA/yum-packaging-precompiled-kmod
 
+%build
+
 # Begin open driver build
 pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}/kernel-open
 
--- a/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
+++ b/packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
@@ -258,7 +258,7 @@ pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}/supported-gpus
 jq -r '.chips[] | select(.features[] | contains("kernelopen")) |
 select(.devid != "0x1DB1"
 and .devid != "0x1DB5"
-and .devid != "0x1DEB8"
+and .devid != "0x1EB8"
 and .devid != "0x1EB4"
 and .devid != "0x2237")' supported-gpus.json | jq -s '{"open-gpu": .}' > open-gpu-supported-devices.json
 # confirm "NVIDIA H100" is in the resulting file to catch shape changes
--- a/packages/kmod-6.1-nvidia-r580/open-gpu-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r580/open-gpu-license-fallback.service
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 RemainAfterExit=true
 
 [Install]
-WantedBy=nvidia-k8s-device-plugin.service
+WantedBy=nvidia-gridd.service
--- a/packages/kmod-6.1-nvidia-r580/tesla-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r580/tesla-license-fallback.service
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 RemainAfterExit=true
 
 [Install]
-WantedBy=nvidia-k8s-device-plugin.service
+WantedBy=nvidia-gridd.service
--- a/packages/kmod-6.1-nvidia-r595/grid-license-check.service
+++ b/packages/kmod-6.1-nvidia-r595/grid-license-check.service
@@ -14,6 +14,3 @@ ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-licen
 ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
 ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 StandardOutput=append:/tmp/.nvidia-gridd-license
-
-[Install]
-WantedBy=nvidia-k8s-device-plugin.service
--- a/packages/kmod-6.1-nvidia-r595/kmod-6.1-nvidia-r595.spec
+++ b/packages/kmod-6.1-nvidia-r595/kmod-6.1-nvidia-r595.spec
@@ -188,6 +188,8 @@ tar -xf %{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
 # This recipe was based in the NVIDIA yum/dnf specs:
 # https://github.com/NVIDIA/yum-packaging-precompiled-kmod
 
+%build
+
 # Begin open driver build
 pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}/kernel-open
 
--- a/packages/kmod-6.1-nvidia-r595/kmod-6.1-nvidia-r595.spec
+++ b/packages/kmod-6.1-nvidia-r595/kmod-6.1-nvidia-r595.spec
@@ -258,7 +258,7 @@ pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}/supported-gpus
 jq -r '.chips[] | select(.features[] | contains("kernelopen")) |
 select(.devid != "0x1DB1"
 and .devid != "0x1DB5"
-and .devid != "0x1DEB8"
+and .devid != "0x1EB8"
 and .devid != "0x1EB4"
 and .devid != "0x2237")' supported-gpus.json | jq -s '{"open-gpu": .}' > open-gpu-supported-devices.json
 # confirm "NVIDIA H100" is in the resulting file to catch shape changes
--- a/packages/kmod-6.1-nvidia-r595/open-gpu-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r595/open-gpu-license-fallback.service
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 RemainAfterExit=true
 
 [Install]
-WantedBy=nvidia-k8s-device-plugin.service
+WantedBy=nvidia-gridd.service
--- a/packages/kmod-6.1-nvidia-r595/tesla-license-fallback.service
+++ b/packages/kmod-6.1-nvidia-r595/tesla-license-fallback.service
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 RemainAfterExit=true
 
 [Install]
-WantedBy=nvidia-k8s-device-plugin.service
+WantedBy=nvidia-gridd.service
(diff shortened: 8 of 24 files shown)


ginglis13 (approved):
LGTM - nit on missing space after `:` in commit message `kmod-nvidia:fallback.service to install on all variants`

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
```diff
@@ -169,6 +169,8 @@ tar -xf %{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
 # This recipe was based in the NVIDIA yum/dnf specs:
 # https://github.com/NVIDIA/yum-packaging-precompiled-kmod
 
+%build
+
```
arnaldo2792, line 192: Two nits:

- I would have kept this commit out of the series, because you are fixing three different problems
- You could have used kmod-*nvidia-* in the commit message

packages/kmod-6.1-nvidia-r580/kmod-6.1-nvidia-r580.spec
```diff
@@ -239,7 +239,7 @@ pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_ver}/supported-gpus
 jq -r '.chips[] | select(.features[] | contains("kernelopen")) |
 select(.devid != "0x1DB1"
 and .devid != "0x1DB5"
-and .devid != "0x1DEB8"
+and .devid != "0x1EB8"
```
arnaldo2792, line 261: In your PR description, can you add the testing you did here to validate that `g4dn.*` are using the correct driver. And add in the commit message that we are following the steps of the EKS Optimized AMI.

packages/kmod-6.1-nvidia-r580/open-gpu-license-fallback.service
```diff
@@ -11,4 +11,4 @@ ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 RemainAfterExit=true
```
arnaldo2792, line 11: The commit message is not well formatted and confusing. Could you please expand on how your change fixes the problem with non-k8s variants?

Also, should the service run `After=nvidia-gridd.service`? `WantedBy` doesn't guarantee ordering.

packages/kmod-6.1-nvidia-r580/grid-license-check.service
```diff
@@ -14,6 +14,3 @@ ExecStart=/usr/bin/grep -q "License Status.*: Licensed" /tmp/.nvidia-gridd-licen
 ExecStart=/usr/bin/touch /etc/drivers/.grid-licensed
 ExecStart=/usr/bin/systemctl stop grid-license-check.timer --no-block
 StandardOutput=append:/tmp/.nvidia-gridd-license
-
-[Install]
-WantedBy=nvidia-k8s-device-plugin.service
```
arnaldo2792, line 19: In the commit message, or even in this file, explain that this unit doesn't need an `Install` section because the service will be started by a timer.

But also, you must guarantee that the service will start as soon as we reach `multi-user.target`. If you validated this, call it out in the PR description. And in the commit message, call out why the wanted ordering is preserved (e.g. `timers.target` run before, in parallel, after `multi-user.target` whichever is the case we want).

packages/kmod-6.1-nvidia-r580/open-gpu-license-fallback.service
vigh-m: Updated the commit message with more details.

The services can start at the same time as `nvidia-gridd.service`. They no-op if they are on an unsupported instance and will exit without failing the boot.

packages/kmod-6.1-nvidia-r580/grid-license-check.service
vigh-m: Updated the commit message with more info  about this.
We don't need to wait for `multiuser.target` for this service. It was already setup to run after `nvidia-gridd.service` which is installed in `preconfigured.target`. This change shouldn't affect the ordering or install since `timers.target` was always going to install this service.

vigh-m:
⬆️ Update following comments

arnaldo2792 (approved):
One more nit in the PR description:

I understand what's changing with your code changes, but what's the behavior your change fixed? I understand the nvidia-gridd.service now starts in all variants, but what problem was this causing? Failed boots? Excessive logging? A service running that shouldn't? Or a service not running that should?

vigh-m:
> One more nit in the PR description:
> 
> I understand what's changing with your code changes, but what's the behavior your change fixed? I understand the nvidia-gridd.service now starts in all variants, but what problem was this causing? Failed boots? Excessive logging? A service running that shouldn't? Or a service not running that should?

Sure. Added more context. I had also linked the Github Issue that caused this code change

maherthomsi (approved):

PR #546, opened 2026-09-03
actions: add kernel config validation

jpculp:
**Description of changes:**

Add a CheckConfig workflow that runs 'make full-config' on both x86_64 and aarch64 GitHub-hosted runners and fails if any tracked kernel config files change.

**Testing done:**

- [x] Ran workflow with good configs and PASS
- [x] Ran workflow with bad configs and FAIL

**Terms of contribution:**

### DIFF
--- a/.github/workflows/check-config.yml
+++ b/.github/workflows/check-config.yml
@@ -1,0 +1,43 @@
+name: CheckConfig
+on:
+  pull_request:
+    branches: [develop]
+    paths:
+      - 'packages/kernel-*/**'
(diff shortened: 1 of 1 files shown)


piyush-jena (approved):

vigh-m (approved):

jpculp:
Just to add some context on why now, I realized we are building both x86_64 and aarch64 on x86_64 runners. This means that there is a non-zero chance of a misconfig for any files specific to builds on aarch64.

ginglis13 (approved):

